PR #5, opened 2023-02-14
Add auto-purge test cases

pasin:
* Added 003-replication-auto-purge test cases.
* Added new posts dataset
* Removed default_empty dataset as not needed


bmeike (approved):
Do you have to edit these files every time you change a test?

pasin:
Yes. I don't think we are going to touch them much in the future besides adding new tests.

PR #11, opened 2023-07-20
Update REST API Spec

pasin:
* /getAllDocumentIDs : changed to /getAllDocuments which will return an object where key is the collection name and the value is an array of objects containing document info including document ID and revisionID.

* /startReplicator : changed ReplicationCollection.collection to names which is an array of collection names to be able to specify multiple collections sharing the same configuration.

* /getReplicatorStatus : changed ReplicatorStatus.progress.complete to completed which is a boolean value indicating that the replication is completed or not and removed ReplicatorStatus.progress.documentCount

* Will have a separate PR for the error code 400 / 500.

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -61,13 +61,13 @@ paths:
           $ref: '#/components/responses/Error'
         '405':
           description: Operation Not Allowed
-  /getAllDocumentIDs:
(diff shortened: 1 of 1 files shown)


bmeike (approved):
excellent

borrrden (approved):

PR #19, opened 2023-08-01
Update /updateDatabase API

pasin:
* Updated /updateDatabase API to use keypaths.
* Bumped version to 0.1.5.

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -139,23 +139,41 @@ paths:
         dictionary properties and array elements. The keypath looks like "foo.bar[1].baz" - The keypath components are 
         separated by dot ('.'). Each path is represented by the key name in a dictionary object, and the index brackets 
         at the end of the key name points to an array element in an array object. A '\' can be used to escape a special 
-        character ('.', '[' or '$') in the key name.
+        character ('.', '[', ']', '\' or '$') in the key name.
 
         The value at the location specified by the keypath is replaced with the value associated with the keypath, in the object 
-        regardless of its type. When a keypath includes a non-existent property (that is, an object that does not contain 
-        the specified key), the key is added and its value set as a dictionary or an array if the key refers to an array element
-        (the key end with array index brackets). 
+        regardless of its type. When a path in the keypath refers to a non-existent property, the key is added and 
+        its value set as a dictionary or an array depending on the type of the path.
 
-        When the keypath includes an array index that is out of bounds (the array is smaller than the specified index) the array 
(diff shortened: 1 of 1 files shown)


spec/api/api.yaml
```diff
@@ -123,8 +126,38 @@ paths:
...
+        When the keypath includes an array index that is out of bounds (the array is smaller than the specified index) the array 
+        is padded with nulls until it is exactly large enough to make the specified index legal.  
+        
+        A keypath that includes an element of the wrong type -- a object when an array index is specified in the path, 
+        an array when object property appears in the path, or any path that crosses a scaler -- is an error.
+
+        removedProperties:
+
+        An array of the keypaths which refer to ditionary keys or array elements to be removed. If a keypath refer to a non-existent
+        property or array element, the error will be returned. When removing multiple items in an array, the order of the array indexes
```
borrrden, line 158: Does it need to be an error if the keypath doesn't exist?  The end result is the same (that keypath has no value)

spec/api/api.yaml
```diff
@@ -123,8 +126,38 @@ paths:
...
+        updatedProperties:
+
+        An array of objects. Each object contains one or more keys that are keypaths into the document to be updated with the 
+        values associated with the keys. The objects in the array are evaluated in their natural order.
+
+        The keypath identifies a unique location in the document's property tree starting from the root that follows 
+        dictionary properties and array elements. The keypath looks like "foo.bar[1].baz" - The keypath components are 
+        separated by dot ('.'). Each path is represented by the key name in a dictionary object, and the index brackets 
+        at the end of the key name points to an array element in an array object. A '\' can be used to escape a special 
+        character ('.', '[' or '$') in the key name.
```
borrrden, line 142: Don't forget `]` as a special character, and `\` itself.

spec/api/api.yaml
```diff
@@ -123,8 +126,38 @@ paths:
...
+
+        The value at the location specified by the keypath is replaced with the value associated with the keypath, in the object 
+        regardless of its type. When a keypath includes a non-existent property (that is, an object that does not contain 
+        the specified key), the key is added and its value set as a dictionary or an array if the key refers to an array element
+        (the key end with array index brackets). 
+
+        When the keypath includes an array index that is out of bounds (the array is smaller than the specified index) the array 
+        is padded with nulls until it is exactly large enough to make the specified index legal.  
+        
+        A keypath that includes an element of the wrong type -- a object when an array index is specified in the path,
```
borrrden, line 152: Slightly confusing wording.  You mean to say when the keypath does not match the existing object?  Perhaps a few examples are in order.

borrrden (approved):

PR #21, opened 2023-08-02
Add KeyPathParser and KeyPath

bmeike:
There's a lot of refactoring, here.  The only part that I care about getting reviewed are KeyPathParser and KeyPath


servers/jak/shared/common/main/java/com/couchbase/lite/mobiletest/data/KeypathParser.java
```diff
@@ -0,0 +1,171 @@
...
+                default:
+                    throw new ClientError(errMsg("Unexpected character " + curChar));
+            }
+        }
+    }
+
+    @NonNull
+    private KeyPath.PathElem.Property parseProperty() {
+        // skip and unescaped leading '$'
+        if (curCharClass == CharClass.DOLLAR) { nextChar(); }
```
borrrden, line 70: I believe the current way dollar sign is supposed to be handled is the same way as JSON paths, meaning that dollar sign indicates the root of the object so it would look like `$.foo.bar`.  This just looks like being able to add an arbitrary dollar sign onto the front?

servers/jak/shared/common/main/java/com/couchbase/lite/mobiletest/data/KeypathParser.java
```diff
@@ -0,0 +1,171 @@
...
+            case "5":
+            case "6":
+            case "7":
+            case "8":
+            case "9":
+                curCharClass = CharClass.DIGIT;
+                return;
+            case "\\":
+                nextCh();
+                if ("".equals(curChar)) { throw new ClientError(errMsg("Dangling escape ('/')")); }
```
borrrden, line 151: Your error message slash is the wrong direction.

pasin (approved):
Nice design that can get value from the tree and modify value from the JSON tree.

servers/jak/shared/common/main/java/com/couchbase/lite/mobiletest/data/KeypathParser.java
bmeike: Yes.  Fixed.  I sent around a grammar for what I've ended up implementing.  I think it is more in line with your ideas.

servers/jak/shared/common/main/java/com/couchbase/lite/mobiletest/data/KeypathParser.java
bmeike: Yep.  Fixed.

PR #48, opened 2023-08-25
API : Update API to support tests with blob

pasin:
* Added updatedBlobs option to /updateDatabase.
* Added verification logic info including blob verification logic in /verifyDocuments.
* Added new /performMaintenance API

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -35,6 +35,12 @@ info:
 
     Changes
 
+    0.2.2 (08/25/23)
+    * Added updatedBlobs option to /updateDatabase.
+    * Added verification logic info including blob verification logic in /verifyDocuments.
+    * Added blob file doesn't exist reason to /verifyDocuments description.
+    * Added new /performMaintenance API
+
     0.2.1 (08/18/23)
     * Added note about out-of-scope functions to /verifyDocuments.
     * Changed wording in /snapshotDocuments description. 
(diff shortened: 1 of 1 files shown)


spec/api/api.yaml
```diff
@@ -369,6 +389,17 @@ paths:
...
 
+        Verification Logic:
+        1. If the document specified in the delta doesn't exist in the snapshot, the 400 request error will be returned.
+        2. If the document specified in the delta with change type PURGE or DELETE, the corresponding snapshot document 
+           value in the snaphot MUST be null.
+        3. If the document specified in the delta with change type UPDATE, the expected document which is the 
+           corresponding snaphot document applied with the specified delta will be compared with the actual document 
+           which is the current document in the collection. The body of expected and actual document MUST be the same. 
+        4. When checking if the two blob objects are equals, the dictoinary properties of the two blobs MUST be the same.
+           In addition, the actual blob file in the database MUST exist; this can be checked by calling Database's
```
bmeike, line 400: I guess this is ok, right?  There is no way for the file's contents to get out of sync with its name?

bmeike (approved):
Looks good to me.  What is the target date for this?

spec/api/api.yaml
pasin: With the recent CBSE bug that blob files in the database could be gone, I would like to check for the file existent as well. If this becomes extensive to test, we could have a separate API to test that.

pasin:
> What is the target date for this?

I'm thinking about next sprint but let discuss this what we have left and if it makes sense or not. On another note, we received the purchased machines (Dan is setting them up for us) so I expect that we will spend sometime working on getting the tests live on Jenkins.

bmeike:
Akkk!!!  Nooooooooooooooo!  ;-P

I have to learn .NET, and there are upcoming releases of all 3 of my products for 2 different versions!  I gotta get some other work done! :-p

PR #64, opened 2024-08-15
Add runQuery endpoint to test server API

borrrden:
Will be used for tests which run queries and compare against couchbase server


bmeike (approved):

bmeike:
I made two suggestions:
- I think that, elsewhere, the key for the database is “database” spelled out in full.
- I suggest that the return from runQuery be:   { “results” : [ … ] }

pasin (approved):

PR #70, opened 2024-09-17
Update APIs as required by SG Team

pasin:
* Updated the request body of the /reset to support creating an empty database with predefined collections.

* Added /stopReplicator.

* Added /getDocument.

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -126,15 +126,22 @@ paths:
         - API
       summary: Reset the test server
       description: |- 
-        Reset the test server by deleting all of the databases and re-creating new databases with prebuilt datasets if specified. 
+        Reset the test server by deleting all of the databases and re-creating new databases as specified. 
       operationId: reset
       requestBody:
         description: |- 
-          The request object describes how the databases will be setup after being reset. The properties include:
+          The request object describes how the databases will be setup after being reset. The request object
+          contains one optional key named "databases" which specifies the databases to be created. If not 
+          specified,  no databases will be created.
 
-          * **datasets** (Optional) : the datasets which are prebuilt databases to be copied as the specified database names.
-            The data structure is a JSON dictionary where the keys are the dataset names and the values are an array of the
-            database names. Without specified, no databases will be created.  The special catalog name 'empty' will create databases with no content inside.
+          The "databases" if specified contains a dictionary describing how the database to be created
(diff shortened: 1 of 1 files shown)


spec/api/api.yaml
```diff
@@ -914,13 +990,25 @@ components:
...
+            will be created.
           type: object
           additionalProperties:
-            type: array
-            items:
-              type: string
-          example: { 'catalog': ['db1', 'db2'] }
+            type: object
+            properties:
+              collections:
```
borrrden, line 1001: You can probably use the "oneOf" keyword to specify that one or the other is used

borrrden (changes requested):

borrrden (changes requested):
Bump API version and add changes to changelog please

borrrden (approved):

spec/api/api.yaml
```diff
@@ -126,15 +131,22 @@ paths:
         - API
       summary: Reset the test server
       description: |- 
-        Reset the test server by deleting all of the databases and re-creating new databases with prebuilt datasets if specified. 
+        Reset the test server by deleting all of the databases and re-creating new databases as specified. 
       operationId: reset
       requestBody:
         description: |- 
-          The request object describes how the databases will be setup after being reset. The properties include:
+          The request object describes how the databases will be setup after being reset. The request object
```
bmeike, line 138: Shouldn't this have a ref to an explicit schema?

spec/api/api.yaml
```diff
@@ -185,6 +197,43 @@ paths:
           $ref: '#/components/responses/Error'
         '500':
           $ref: '#/components/responses/Error'
+  /getDocument:
```
bmeike, line 200: Gosh I'm sorry to see this go in.  I predict that it will come back to haunt us.

... we should discuss getting rid of the whole server-side snapshot/compare facility...

bmeike (approved):

spec/api/api.yaml
pasin: This is the simplest solution compared to the others.

spec/api/api.yaml
pasin: It has a reference to an explicit schema. I could move this description there.

pasin: Done.

PR #72, opened 2024-09-26
Proposed newSession endpoint

bmeike:
My proposal for extending the /setupLogging endpoint as /newSession.

Of note:
- This will require a change to all of python tests.  I think it will be easy: each suite needs to start a session
- I don't think that the example given for the log slurper URL is actually a URL.  Should it be?
- One of the editors has removed all the extraneous spaces.  I suggest viewing difference with the &w=1 flag.

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -34,6 +34,8 @@ info:
       ESCAPE   ::= '\' (! EOP )
 
     Changes
+    0.6.0 (09/27/2025)
+    * Replace setupLogging endpoint with expanded newSession endpoint
 
     0.5.2 (09/25/2025)
     * Split getDocument into database and document key to more closely match other API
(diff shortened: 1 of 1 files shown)


borrrden (approved):

borrrden:
The example for the log slurper could possibly be changed to include a prefix but it would be different depending on the operations (some are http and some are ws)

borrrden (changes requested):
https://github.com/couchbaselabs/couchbase-lite-tests/blob/main/spec/api/api.yaml#L108

Needs a bump to 0.6.0

bmeike:
@borrrden : huh?  It is 0.6.0!!

bmeike:
I see.  There is a comment for the version and the actual version.  FIxed

PR #83, opened 2024-11-01
Separate Java Desktop and WS tests

bmeike:
Passes validations

### DIFF
--- a/jenkins/pipelines/java/Jenkinsfile
+++ b/jenkins/pipelines/java/Jenkinsfile
@@ -6,7 +6,6 @@ pipeline {
         string(name: 'CBL_BUILD', defaultValue: '', description: 'Couchbase Lite Build Number')
         string(name: 'SGW_URL', defaultValue: '', description: "The url of Sync Gateway to download")
     }
-    options { timeout(time: 200, unit: 'MINUTES') }
     stages {
         stage('Init') {
             steps {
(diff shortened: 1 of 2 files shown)


jenkins/pipelines/java/Jenkinsfile
```diff
@@ -17,17 +16,24 @@ pipeline {
                 }
             }
         }
-        stage('Run Tests') {
+
+        // Note that there is no fundamental reason that the Desktop and Web Services
+        // builds couldn't run in parallel.  We'd need to figure out how to set
+        // up the environment server and sg at alternate ports, probably, and we'd
+        // need hardware that can support the load.
```
borrrden, line 23: This note is exactly why I realized that ultimately making device specific locks is redundant.  Even if it's two different phones running servers, they will be talking to the same backend until we migrate to AWS.

jenkins/pipelines/java/Jenkinsfile
```diff
@@ -17,17 +16,24 @@ pipeline {
...
+
+        stage('Run Desktop Tests') {
             parallel {
                 stage('OSX Tests') {
                     agent { label 'mob-e2e-mac-01' }
+                    options {
+                        timeout(time: 45, unit: 'MINUTES')
+                        lock("mob-e2e-mac-01")
```
borrrden, line 31: This is not going to work unless you've set up a lockable resource on Jenkins.  It's probably confusing but I named them the same as their device name to make it clear what they are.

jenkins/pipelines/java/Jenkinsfile
```diff
@@ -41,29 +47,18 @@ pipeline {
...
-                                }
-                            }
-                        }
                     }
                 }
                 stage('Windows Tests') {
                     agent { label 'mob-e2e-win-01' }
+                    options {
+                        timeout(time: 45, unit: 'MINUTES')
+                        lock("mob-e2e-win-01")
```
borrrden, line 56: Ditto here for not working

jenkins/pipelines/java/Jenkinsfile
```diff
@@ -77,45 +72,97 @@ pipeline {
...
                         }
-                        stage('Windows Web Services Tests') {
-                            when { expression { false } }
+                    }
+                }
+                stage('Linux Tests') {
+                    agent { label 'mob-e2e-deb-02' }
+                    options {
+                        timeout(time: 45, unit: 'MINUTES')
+                        lock("mob-e2e-deb-02")
```
borrrden, line 81: Also once again

jenkins/pipelines/java/Jenkinsfile
```diff
@@ -77,45 +72,97 @@ pipeline {
...
                     }
                 }
-                stage('Linux Tests') {
-                    agent { label 'mob-e2e-deb-02' }
+            }
+        }
+        stage('Run Webservices Tests') {
+            parallel {
+                stage('OSX Tests') {
+                    agent { label 'mob-e2e-mac-01' }
```
borrrden, line 107: In fact these locks won't actually protect against anythiing at the moment since all your jobs that use the same machine run seqentially right?

borrrden (changes requested):

jenkins/pipelines/java/Jenkinsfile
bmeike: So, in order ;-P
- All of those things that I am locking are lockable resources.  I, too, named them for the devices that host the task
- I'm not running tasks on the same device, in parallel, it is true.  These locks are meant to protect my jobs against YOUR jobs.  Of course that won't work unless you start using the locks, too.  Please do?

PR #91, opened 2024-11-12
Change .NET test server to use port 5555

borrrden:
To avoid accidentally using another accidentally running test server from another platform


pasin (approved):
Should each platform use a different port as well?

borrrden:
The first time I ever ran the test server in the E2E lab, the Java Android test server had been left running and I was unable to start my server without killing it, so that makes it a real concern for me.

PR #93, opened 2024-11-14
Update SG config and dataset to workaround SG 4.0

pasin:
* In SG bootstrap config, don’t use TLS for connection with CBS
* In SG config for datasource, disable import docs.


borrrden (commented):
What about posts-config ?

pasin:
It didn't get push somehow.

PR #94, opened 2024-11-21
Update TDK client to handle SG added cv key

bmeike:
Also:
- noted in the API spec that the rev key in the response to /getAllDocuments may be either a revId or a CV
- added some some build notes
- Effing PyCharm decided to reformat some stuff

The actual changes are in syncgateway.py.  I recommend appending the 'w=1' flag to the end of the code URL so that github's comparison ignores changes in whitespace

Basic replication tests seem to pass using CBL-Java 3.2.1


borrrden (approved):

pasin (approved):
Look good to me. The only thing that I don't know if we need to add an additional parameter to the SG REST API to enable this cv feature.

bmeike:
We do need that additional parameter: `_all_docs?show_cv=true`

Added and typing fixed.

PR #95, opened 2024-11-21
Implement remote logger in iOS Test Server

pasin:
* Implemented remote logger in iOS Test Server
* Installed remote logger in newSession request’s handler.
* Refactored the code to keep session manager and database manager  in Application’s shared storage.
* Moved TestServer.logger which is a console logger to Log class as ConsoleLogger.


borrrden (approved):
Looks ok if I ignore all the changes related to session only.

PR #100, opened 2024-11-23
Fix missing role error in test-fest tests

pasin:
* Problem: The sync function for the list docs assigned the contributor role to the user and the contributor is required for adding tasks to the tasks channel. When pushing both list and its task docs to SG at the same, there is no order guranteed that the list doc will be process and the role will assign to the user first. As a result, sometime, the sync function rejects the task doc as the role is missing.

* Update the tests to wait for the list doc to be successfully pushed after creating the list doc.


pasin:
Will fix this in a different way.

PR #101, opened 2024-11-23
CBL-6495: Fix missing roles when pushing tasks in test-fest tests

pasin:
* Problem: The contributor role required by the task in the tasks' sync function is assigned to the user when the list is pushed in lists' sync function. However, the list doc and its task docs are pushed independenly and there is no order guaranteed that the list will be pushed first. As a result, if the task doc is pushed first,  the task doc will be rejected by the sync function as the contributor for the user is missing.

* Instead of using sync function to assign the contributor role to the user, assign the required roles to the users using SG REST API at the beginning of the test where the roles are being created now.

* Reorder some steps in tests to be inline with the test spec.

### DIFF
--- a/tests/test_fest.py
+++ b/tests/test_fest.py
@@ -13,23 +13,32 @@ from cbltest.api.replicator_types import ReplicatorBasicAuthenticator, WaitForDo
 
 
 class TestFest(CBLTestClass):
-    async def setup_test_fest_cloud(self, cblpytest: CBLPyTest, dataset_path: Path,
-                                    roles: Optional[List[str]] = None) -> CouchbaseCloud:
+    async def setup_test_fest_cloud(self, cblpytest: CBLPyTest, dataset_path: Path, 
+                                    roles: Optional[Dict[str, List[str]]] = None) -> CouchbaseCloud:
+        """
+        Sets up the test environment.
+        :param cblpytest: CBLPyTest object.
+        :param dataset_path: Path to the SG dataset.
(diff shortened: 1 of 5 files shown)


bmeike (approved):
Quite a big change!

tests/test_fest.py
```diff
@@ -164,6 +176,10 @@ async def test_update_task(self, cblpytest: CBLPyTest, dataset_path: Path) -> No
             DocumentEntry("_default.tasks", "db1-list1-task1")
         ])
 
+        self.mark_test_step("Start the replicators")
```
borrrden, line 179: This should be updated in the header and test spec as one of the steps.

tests/test_fest.py
```diff
@@ -199,6 +205,12 @@ async def test_update_task(self, cblpytest: CBLPyTest, dataset_path: Path) -> No
         verify_result = await db2.verify_documents(snapshot_updater)
         assert verify_result.result is True, f"Unexpected docs in db2: {verify_result.description}"
 
+        self.mark_test_step("Snapshot db1")
```
borrrden, line 208: Also this should be updated in the test spec.

borrrden (approved):
Could use some cleanup with test spec vs marked test steps.  Some of the ones I marked may already be valid, but I expect to be able to find each test step from the test spec in the test output.

tests/test_fest.py
pasin: I have changed this based on the test spec.

pasin:
Agreed, I'm working on it.

PR #112, opened 2024-12-17
Update main and iOS script & pipeline to support dataset version

pasin:
Updated iOS script & pipeline to support dataset version. Updated main pipeline to support dataset version.


bmeike (approved):
This look very familiar.
I'm interested that both you and Jim copy the lib over, instead of just pointing at it.

pasin:
I guess I can create a symlink to the selected folder. Will try that to see if it works with Xcode project for iOS or not.

borrrden:
.NET usually has issues with doing that kind of thing

PR #114, opened 2024-12-19
Enforce automatic formatting.

torcolvin:
I'm using ruff but black would be fine.

There are several commits, which are generally automated:

- run `ruff format`
- run `ruff check --fix`
- add import sorting to `pyproject.toml` and run import again
- make sure that mypy continues to work, explicitly ignore couchbase and opentelemetry which weren't declared with types

I don't care about mypy much at all, and don't really care about sorting imports or even fixing the linting issues, but I thought enforcing some consistency is helpful.

What is difficult is that without autoformatting, everytime I edit a file for an upcoming fix, that I accidentally change the contents of existing files because all my editors run with automatic formatting.


tests/test_replication_filter_data.py
```diff
@@ -318,5 +318,5 @@
     "landmark_897",
     "landmark_898",
     "landmark_899",
-    "landmark_900"
```
borrrden, line 321: Why in the world would this be "better"?

borrrden (approved):

tests/test_replication_filter_data.py
torcolvin: These are the opinions of the community, with the idea that you don't get a choice in the style: https://github.com/psf/black#the-black-code-style

However, this is preferred because python shares the "feature" of C where two strings next to each other get concatenated, and it prevents you appending the set with `landmark901` and then accidentally having the last entry as `landmark900landmark901`:

```
foo = { "landmark1",
     ...
     "landmark900"
     "landmark901"
}
```

Perhaps the question to ask is why JSON does not allow trailing commas?

PR #120, opened 2025-01-13
Android: Failed tests should fail the Jenkins job

bmeike:


pasin (approved):
I think I need to do the same for C and iOS.

PR #126, opened 2025-01-21
Refactor now that nameof is gone from varname

borrrden:
The issues on the repo indicate to re-implement the method in terms of another one of the methods inside the package (as I did in utils.py)


bmeike (approved):
ugg.

PR #132, opened 2025-01-23
Fix iOS test script and add dataset version to build desc

pasin:
* Fixed swring SGW_URL in iOS test script.
* Added dataset version to build desc of iOS and C.


jenkins/pipelines/ios/test.sh
```diff
@@ -4,7 +4,7 @@ EDITION=${1}
 CBL_VERSION=${2}
 CBL_BLD_NUM=${3}
 CBL_DATASET_VERSION=${4}
-SGW_URL=${4}
+SGW_URL=${5}
```
bmeike, line 7: Whoops! ;-P

PR #135, opened 2025-02-19
CBL-6661: AWS backend scripting support

borrrden:


borrrden:
Note nothing functionally has changed yet, this just provides the ability to change.

environment/aws/sgw_setup/cert/sg_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
bmeike, line 1: Spider sense tingling.... should we be checking in private keys???

bmeike (approved):
This is great.
I wish, though, that you would lead a session in which we all do it once.  This is a huge commit and even with your nice documentation, it is going to require familiarity for others to actually use it.  We need a "platform day" for this.

environment/aws/sgw_setup/cert/sg_key.pem
borrrden: This might set off some alerts, but this is purely for running Sync Gateway with TLS enabled.  I guess I could change this so that it gets generated freshly every time

pasin (approved):

PR #136, opened 2025-03-01
Provide the ability to deploy logslurp to AWS backend

borrrden:
This eliminates any backend dependency on our lab machines.  However, it also means an additional setup step in order to reliably be able to deploy docker onto a remote host.


bmeike (approved):
I'm not up to speed on how this, so I can't really comment on how it works.
Fingers crossed that the new docs explain it all.

borrrden:
The TL;DR is that it is mostly the same docker commands, just utilizing a feature of docker that lets you use a remote machine to do the building and deploying instead of the same one you are running on.  I still plan on having an information session on this, when we actually have a minute...

PR #137, opened 2025-03-06
CBL-6808 CBL6809: Methods for start and stop listener

borrrden:


pasin (approved):
I think we should add a property to the ReplicatorConfiguration for accepting only self-sign certificate so that the test can use that instead of going through the steps to get the cert and pin at least for the starting point.

borrrden:
Alternatively could we just set it that way by default all the time?  Do we have any tests that are NOT using self signed certs?

borrrden:
Cert pinning overrides that property.  It has the highest precedent, so if we make the change nothing will happen and we can start removing the lines in the TDK tests that pin.

PR #138, opened 2025-03-08
Add TDK start and stop listener functionality

borrrden:


bmeike:
Can we get a design review and quick discussion of the implications for the test servers, before this goes in?

bmeike (changes requested):
I imagine this is all good... but it is yet another significant change to the E2E test system that did not get discussed in advance.  Can we just have, if not a quick meeting, then some kind of doc describing this new feature and how it works?

borrrden:
I added it to the discussions meeting agenda

bmeike (approved):
Awesome.

PR #141, opened 2025-03-18
Fixes to get the new aws flow working

borrrden:
For .NET on Jenkins


borrrden:
This has a lot of commits, but the changes are not as wild as the last PR:

- I optimized the deployment of logslurp to reuse existing containers if it finds them
- I introduced a "tag" to the topology json schema so that we can de-depulicate docker contexts.  This will be important if two jobs are running at once so that they don't write over each other's docker context (aka the thing that says "run the docker commands on this machine instead of the local one").  I assume that it will be useful in other places later too.
- Hugely simplified the terraform plan so that it reads the needed pre-existing infra that is reusable from AWS, and only creates the things that need to be scoped to a given test
- Force UTF-8 output for python so that it doesn't fall back to a stone age ascii encoding in which any character above 255 confuses it (especially on Windows)
- Fix the iOS PID detection because setup and teardown happen in two different contexts.  Need to persist the PID of the running test server in a file
- In the diff view on github, all files below topology_schema (except setup_test.py) are specific to .NET and moving .NET to use AWS backend

bmeike (approved):
Probably fine.  It is too big to read and there is no design to which to compare it.

PR #142, opened 2025-03-21
Port remaining QE query tests

borrrden:
Also enhance json_equivalent to print some information about what comparison failed


sanjivanipatrax (approved):

bmeike (approved):
Pretty cool that, at this point, they see to be easy to import.

Just curious if there is a list of tests, somewhere, now?  This seems like a great opportunity to amass a giant spreadsheet o' tests, so that we could, more easily, evaluate coverage....

PR #143, opened 2025-03-21
Fill in the real device ID now that it is available

borrrden:


bmeike (approved):
What is this?

borrrden:
Answered offline

PR #146, opened 2025-03-26
Port replication blob tests from mobile-testkit

borrrden:
The two tests are so similar that their logic is combined into one test


bmeike (approved):
love it

PR #147, opened 2025-03-28
Move the dev tests to the dev_e2e subfolder

borrrden:
Now it can be scoped away from other tests such as QE's


pasin (approved):

borrrden:
> Presumably there is now a tests/qe?

There will be other folders later when QE makes a PR

PR #151, opened 2025-04-01
Implement ability to prebuild and use prebuilt Swift

borrrden:


pasin (approved):

borrrden:
Copying the dataset will come after the dataset becomes a mandatory parameter in a later PR

PR #153, opened 2025-04-02
Make dataset_version mandatory for topology

borrrden:
It is now required in the build_test_server script, and topology.json.  

Modify the builds to copy the dataset

Put the dataset version into the latestbuilds path


bmeike (approved):
that is an amazing amount of change for such a small addition!

PR #156, opened 2025-04-08
Load balancer support, SGW version support

borrrden:
Clean up ugly diffs that always distract me by adding .gitignore


environment/aws/lb_setup/setup_load_balancers.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+
+"""
+This module sets up the LogSlurp service on an AWS EC2 instance. It includes functions for executing remote commands,
```
pasin, line 4: This comment is about LogSlurp.

pasin (approved):

environment/aws/lb_setup/setup_load_balancers.py
borrrden: Good catch, fixed.

PR #157, opened 2025-04-09
Correct .NET pipeline with regards to dataset_version

borrrden:
It needs to actually use it


bmeike (commented):
I will update the test server to use the "blob" directory either end of this week, or early next.

bmeike (approved):

PR #158, opened 2025-04-09
Move the majority of the things in conftest.py into plugins

borrrden:
That way they don't need to be copied and pasted into each test suite


bmeike (approved):
I just don't have bandwidth to actually look at this.
You are moving *very* fast.

borrrden:
I know, it's helpful that I am only working on this and nothing else at the moment but it has that speed side effect.  This doesn't change any behavior, it just makes new test directories easier to create.

PR #161, opened 2025-04-11
Migrate swift ios and java android to AWS / topology

borrrden:


bmeike (approved):
I totally cannot read this.
We are going to need to add this to our "platform day" series.  We are going to need bus insurance for this large, complex system.

borrrden:
I realized a way to simplify this a bit more in another branch so I will do that so that the setup_test.py per platform is quite short.

PR #164, opened 2025-04-14
CBL-6907: add Start and Stop Listener API

bmeike:
Also, some thread safety in TestContext

This starts and stops a listener.  Not clear that it is the one that the tests want: may need to be tweaked....


servers/jak/android/app/build.gradle
```diff
@@ -126,7 +130,10 @@ android {
 repositories {
     if (USE_LOCAL_MAVEN) { mavenLocal() }
     else {
-        maven { url 'https://proget.sc.couchbase.com/maven2/internalmaven' }
+        maven {
+            url 'https://proget.sc.couchbase.com/maven2/internalmaven'
+            content { includeGroupByRegex "com\\.couchbase\\.lite.*" }
```
borrrden, line 135: Ohhh this is an interesting option I like!

borrrden (approved):

PR #165, opened 2025-04-15
Migrate C and Java to AWS / Topology backend

borrrden:
Also refactor and heavily simplify what each platform has to do by moving most of the functionality to a shared setup_test function.


pasin (approved):
Thank you for fixing the C build.

PR #166, opened 2025-04-17
Ensure all jobs are locking correctly and timing out

borrrden:
I noticed that putting the timeout in the stage doesn't give the post section enough time to run in the event of a timeout.  However, that's not desired so I switched the steps to have independent timeouts.

Also the main Jenkinsfile had a syntax error so correct that.


bmeike (approved):
Huh.  One I can read.
Looks good.

PR #167, opened 2025-04-18
Fixes to support prebuilding test servers

borrrden:


borrrden:
This may not be completely functional yet, it needs more testing, but the build and upload part appears to work correctly.  I also made a pretty obvious realization while doing this that there is zero difference in how to build on any desktop platform with Java that needs to be reflected in a topology platform type.  They all build the same gradlew target and the output is the same as well.  So I reduced, e.g. jak_windows_desktop, jak_macos_desktop, and jak_linux_desktop to just jak_desktop.

PR #178, opened 2025-04-30
Move tests specs and pipelines to subfolder

borrrden:
This way other teams can operate in their own folders.  The vast majority of this is moving files into a subfolder and that can cause this to look bigger than it actually is.

I'm tired of changing 100 different relative paths, so make a shared config file to predefine some important locations.

Also there was a small bug in C windows test server build resulting in cblite.dll not getting copied.  There was also a bug in the Android script which I think means that the android sdk install was not properly being run.  

I also deleted the old defunct scripts from all platform folders.


jenkins/pipelines/dev_e2e/android/topology_single_device.json
```diff
@@ -2,7 +2,7 @@
     "$schema": "../../../../environment/aws/topology_setup/topology_schema.json",
     "test_servers": [
         {
-            "platform": "c_android",
```
borrrden, line 5: The renaming algorithm on GitHub got this one wrong, These are two different files.

bmeike (approved):
ugg...

borrrden:
I know I keep dreaming up ways to make large PRs 😛 .  I'm hoping soon I won't have to do that anymore.

PR #179, opened 2025-04-30
Similar to syncgateway.py, created a replication url for CBL DBs

vipbhardwaj:


client/src/cbltest/api/testserver.py
```diff
@@ -12,6 +13,11 @@
     PostResetRequestBody,
 )
 from cbltest.version import VERSION
+from client.src.cbltest import _assert_not_null
```
torcolvin, line 16: This assertion is in `cbltest.assertions`

you can not add the `client` directory, since this gets installed with:

```
cd tests
python3 -m venv venv
. venv/bin/activate
pip install -r requirements.txt
```

and the install step only packages src/cbltest directory as per https://github.com/couchbaselabs/couchbase-lite-tests/blob/main/client/pyproject.toml#L81

vipbhardwaj: Resolved the import, can you please verify the same

PR #183, opened 2025-05-03
Add in a method to skip a test based on platform 

borrrden:
This will help for things like encrypted properties, which are only supported on C, for example.

Also, Fix missed path update from previous PR.

### DIFF
--- a/client/src/cbltest/api/testserver.py
+++ b/client/src/cbltest/api/testserver.py
@@ -1,4 +1,3 @@
-from enum import Flag, auto
 from typing import cast
 
 from opentelemetry.trace import get_tracer
@@ -13,23 +12,6 @@ from cbltest.v1.requests import (
     PostResetRequestBody,
 )
 from cbltest.version import VERSION
-
-
-class ServerVariant(Flag):
-    ANDROID = auto()
-    C = auto()
-    DOTNET = auto()
-    IOS = auto()
-    JAVA = auto()
-
-    def __str__(self) -> str:
-        return "|".join(
-            [
-                member.name
-                for member in ServerVariant
-                if member in self and member.name is not None
-            ]
-        )
 
 
 class TestServer:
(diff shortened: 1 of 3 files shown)


client/src/cbltest/api/testserver.py
```diff
@@ -14,6 +15,23 @@
 from cbltest.version import VERSION
 
 
+class ServerVariant(Flag):
+    ANDROID = auto()
+    C = auto()
+    DOTNET = auto()
+    IOS = auto()
+    JAVA = auto()
```
bmeike, line 23: FWIW, We should, actually, be calling this "JVM".  There are multiple languages (notably Kotlin) that run on the JVM and Java runs on Android.  The distinction between the two things is that one targets ART (on Android) and the other targets the JVM (desktop or webservices).
Perhaps, because this matches our inaccurate naming of the platforms, it is the right thing to do.  Just FYI.
I am entirely willing to change the identifier in the header, if that will help.

client/src/cbltest/api/testserver.py
```diff
@@ -117,3 +144,26 @@ async def log(self, msg: str) -> None:
             TestServerRequestType.LOG, payload
         )
         await self.__request_factory.send_request(self.__index, request)
+
+    def __extract_variant(self, info: GetRootResponse) -> ServerVariant:
```
bmeike, line 148: Is it not possible to make this a method on GetRootResponse?

bmeike (approved):

client/src/cbltest/api/testserver.py
borrrden: I can call it JVM if that will help things.

client/src/cbltest/api/testserver.py
borrrden: Yes, I refactored based on this.

PR #185, opened 2025-05-09
CBL-6117: Add in a test for encrypted properties

borrrden:
- Added in encryptor and decryptor functions for C test server
- Added in TDK python API class for creating encrypted values
- Added a test spec and test for pushing to SGW (skipped if not running C)

### DIFF
--- a/servers/c/src/cbl/CBLManager.cpp
+++ b/servers/c/src/cbl/CBLManager.cpp
@@ -28,6 +28,28 @@ using namespace ts::support::error;
 #define ASSET_BLOBS_DIR "blobs"
 
 namespace ts::cbl {
+    static FLSliceResult xor_cipher(FLSlice input) {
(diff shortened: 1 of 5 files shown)


servers/c/src/cbl/CBLManager.cpp
```diff
@@ -28,6 +28,28 @@ using namespace ts::support::error;
 #define ASSET_BLOBS_DIR "blobs"
 
 namespace ts::cbl {
+    static FLSliceResult xor_cipher(FLSlice input) {
```
bmeike, line 31: Ugg... what is this doing?

bmeike (approved):

pasin (approved):

servers/c/src/cbl/CBLManager.cpp
borrrden: Just something that turns the property into anything but plaintext

PR #188, opened 2025-05-10
Don't ping link local broadcast

borrrden:
And don't fail the job immediately if the ping failed, the ARP entry may still exist anyway


vipbhardwaj (approved):
LGTM

PR #189, opened 2025-05-15
Add migrated tests and infra to run on uberjenkins

vipbhardwaj:
This commit has the first set of migrated tests, as well as the infra needed to run it on uberjenkins, which is in compliance with the directory and file structure being used by dev_e2e.


pasin (commented):
Look good in general. The dev tests also have the test spec for each test. The test spec allows the other to review your test plan (ideally before writing the tests) and to provide a bit info about the purpose of the test. The test spec would also allow the test developers to have a second thought whether the test is really needed or maybe duplicate with the other test. I think it's a good idea for QE to do that as well.

vipbhardwaj:
I have added the specs @pasin , along with a couple of more migrated tests

vipbhardwaj:
I have added a few more tests and the specs related to them.
This commit also contains the new XL blob and the spec for that.
It also contains a new dataset and its config, named short_expiry-sg.json and short_expiry-sg-config.json, and has the spec for that as well

spec/tests/QE/test_delta_sync.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Steps
+1. Reset Sync Gateway and load the `travel` dataset with delta sync enabled.
+2. Reset the local database and load the `travel` dataset.
+3. Start a replicator and perform initial pull replication to ensure all documents are present.
+4. Modify a document in Couchbase Lite.
+5. Start push-pull replication and verify the document is updated in Sync Gateway.
+6. Update documents in Sync Gateway (with and without attachment).
+7. Start push-pull replication and verify only the updated documents are processed due to delta sync.
+8. Check delta sync stats for number of docs updated.
```
borrrden, line 16: I don't understand the significance here.  delta sync is going to save number of bytes, not number of documents.  Even without delta sync only changed documents are processed.

spec/tests/QE/test_delta_sync.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Description
+Verify delta sync behavior within and after the delta revision expiry window. Ensures that deltas are used before expiry and full docs are sent after expiry.
+
+### Steps
+1. Reset Sync Gateway and load a dataset with short delta sync expiry.
+2. Create and update documents in Couchbase Lite.
+3. Replicate to Sync Gateway and record bytes transferred.
+4. Update the document in Sync Gateway.
+5. Wait for the delta revision to expire.
```
borrrden, line 94: How long is this going to take?  I don't want tests waiting around arbitrarily for a long time.

spec/tests/QE/test_no_conflicts.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Description
+Test concurrent updates to the same document in both Sync Gateway and Couchbase Lite, and verify that conflicts are resolved and changes are replicated correctly.
+
+### Steps
+1. Reset Sync Gateway and load the `posts` dataset.
+2. Reset the local database and load the `posts` dataset.
+3. Perform a pull replication from SGW to CBL.
+4. Add a new document in SGW and replicate to CBL.
+5. Update the same document in both SGW and CBL concurrently.
```
borrrden, line 13: Concurrency doesn't matter here since no replication is running, so the word "concurrently" here can be removed

spec/tests/QE/test_no_conflicts.md
```diff
@@ -0,0 +1,47 @@
...
+6. Start a continuous push-pull replication between SGW and CBL.
+7. Verify that the document is updated and conflicts are resolved as expected in both SGW and CBL.
+8. Update the document again in CBL and verify the update is replicated to SGW.
+
+## #2 test_multiple_cbls_updates_concurrently_with_push
+
+### Description
+Test concurrent updates and replication across three Couchbase Lite databases, simulating a multi-device scenario, and verify that all changes are synchronized and conflicts are resolved.
```
borrrden, line 21: A multi device scenario no longer needs to be simulated!

spec/tests/QE/test_replication_eventing.md
```diff
@@ -0,0 +1,26 @@
...
+Test push replication of a large document (>20MB) from Couchbase Lite to Sync Gateway, and verify that an error is thrown for oversized documents.
+
+### Steps
+1. Reset Sync Gateway and load the `posts` dataset.
+2. Reset the local database and load the `posts` dataset.
+3. Create a large document in Couchbase Lite.
+4. Attempt to replicate the large document to Sync Gateway using push replication.
+5. Verify that an error is thrown and the document is not replicated to Sync Gateway.
+
+## #2 test_delta_sync_verification
```
borrrden, line 15: This test looks redundant with test_delta_sync.  What is unique here?

spec/tests/QE/test_replicator_encryption_hook.md
```diff
@@ -0,0 +1,39 @@
...
+1. Reset Sync Gateway and load the `posts` dataset.
+2. Reset the local database and load the `posts` dataset.
+3. Replicate from SGW to CBL and verify initial documents.
+4. Create a document in CBL with an encrypted value at the 15th level of nesting.
+5. Replicate to SGW and verify the encrypted field is present in SGW.
+
+## #2 test_delta_sync_with_encryption
+
+### Description
+Verify that delta sync does not work when an encryption callback hook is present, and that documents with encrypted fields are not editable in SGW.
```
borrrden, line 18: Which part asserts the last part of this sentence?

spec/tests/QE/test_replicator_encryption_hook.md
```diff
@@ -0,0 +1,39 @@
...
+### Description
+Verify push/pull replication works with large data, including updates with and without attachments, and that delta sync stats show bandwidth saving and correct document processing.
+
+### Steps
+1. Reset Sync Gateway and load the `travel` dataset with delta sync enabled.
+2. Reset the local database and load the `travel` dataset.
+3. Start a replicator and perform initial replication.
+4. Modify documents in CBL and replicate changes.
+5. Update documents in SGW with and without attachments.
+6. Perform push/pull replication and verify only updated documents are processed.
```
borrrden, line 39: Again this is weird in that "only updated documents are processed" is a universal truth not specific to delta sync

tests/QE/config.example.json
```diff
@@ -0,0 +1,7 @@
+{
```
borrrden, line 1: This file is probably unneeded

tests/QE/test_replication_eventing.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        await cloud.configure_dataset(dataset_path, "posts")
+
+        self.mark_test_step("Reset local database, and load `posts` dataset")
+        dbs = await cblpytest.test_servers[0].create_and_reset_db(
+            ["db1"], dataset="posts"
+        )
+        db = dbs[0]
+
+        self.mark_test_step("Start replication")
```
borrrden, line 46: This step is not in the spec at this point.  The spec and these marked test steps should be identical!

borrrden (changes requested):
I will start with the spec and other things before getting to the actual test implementation.  Also please rebase to get a clean GH actions run.

tests/QE/test_replication_eventing.py
vipbhardwaj: Oh, okay, I'll update the specs with the test steps to be "identical" with the `mark_test_step` statements.

client/src/cbltest/api/syncgateway.py
```diff
@@ -427,6 +427,12 @@ def replication_url(self, db_name: str):
         _assert_not_null(db_name, "db_name")
         return urljoin(self.__replication_url, db_name)
 
+    def get_expvars(self):
```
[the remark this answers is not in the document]
vipbhardwaj: Oh, right, so I intended on adding a method related to `get_expvars` which is right now being used from inside the test file to rather be in a base file like so, but I couldn't get to the same yet.
Its a good point though, shall I keep the method in the delta_sync test file itself or in a base file? Where do you say its better suited?

spec/tests/QE/test_delta_sync.md
vipbhardwaj: You're right, that's an un-rectified statement from my side earlier in the month when I started with these tests.

spec/tests/QE/test_replication_eventing.md
vipbhardwaj: There is no test named : `test_delta_sync`, as in the current branch I checked right now. Is it something added recently in some other branch?
If so, I can remove this redundant test.

spec/tests/QE/test_replicator_encryption_hook.md
vipbhardwaj: That was in the description of testkit while I was migrating it early on, went back to check and realised for some reason they did not emphasise on that at all as well.
Removed that misleading step from the description of TDK test as well as from the spec. ✅

tests/QE/config.example.json
vipbhardwaj: Done, please mark resolved if seem correct

spec/tests/QE/test_replication_eventing.md
vipbhardwaj: I realised and fixed the issue, please mark resolved if seem correct

jenkins/pipelines/QE/ios/Jenkinsfile
```diff
@@ -0,0 +1,61 @@
...
+            environment {
+                KEYCHAIN_PASSWORD = credentials('mobile-qe-keychain')
+            }
+            steps {
+                // Unlock keychain:
+                sh 'security unlock-keychain -p ${KEYCHAIN_PASSWORD} ~/Library/Keychains/login.keychain-db'
+                echo "Run iOS Test"
+                timeout(time: 60, unit: 'MINUTES') {
+                    sh """
+                        export PATH="/opt/homebrew/opt/python@3.10/bin:/opt/homebrew/bin:/usr/local/bin:\$PATH"
```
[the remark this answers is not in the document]
vipbhardwaj: Done, please mark resolved if seem correct

dataset/sg/short_expiry-sg-config.json
```diff
@@ -0,0 +1,42 @@
...
+    "delta_sync": {
+      "delta_sync": {
+        "enabled": true,
+        "rev_max_age_seconds": 60
+      }
+    },
+    "delta_sync_with_expiry": {
+      "delta_sync": {
+        "enabled": true,
+        "rev_max_age_seconds": 60
```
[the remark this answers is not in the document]
vipbhardwaj: Removed it, please mark resolved if seem correct

spec/tests/QE/test_delta_sync.md
vipbhardwaj: Done, please mark resolved if seem correct

spec/tests/QE/test_replicator_encryption_hook.md
vipbhardwaj: Done, please mark resolved if seem correct

client/src/cbltest/api/syncgateway.py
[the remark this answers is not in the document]
borrrden: The naming of it suggested that it was meant to pull down the expvars values, not simply construct its URL.  That's why it looked odd to me.

spec/tests/QE/test_no_conflicts.md
borrrden: The spec says to simulate a multi device scenario.  We HAVE multiple devices now, so why not use multiple devices?

spec/tests/QE/test_no_conflicts.md
```diff
@@ -0,0 +1,60 @@
...
+Test concurrent updates to the same document in both Sync Gateway and Couchbase Lite, and verify that conflicts are resolved and changes are replicated correctly.
+
+### Steps
+1. Reset SG and load `posts` dataset
+2. Reset local database and load `posts` dataset
+3. Pull replication (continuous) to CBL
+4. Wait until the replicator is idle
+5. Create docs in SG
+6. Update docs in SGW and CBL
+7. Start Push Pull replication between SGW and CBL (continuous)
```
borrrden, line 15: You are going to start another continuous replication while the previous continuous replication is running?

spec/tests/QE/test_replicator_encryption_hook.md
borrrden: I still see "documents with encrypted fields are not editable in SGW".  First of all I don't understand what that means, exactly.  They can still be edited, they just can't be viewed.

tests/QE/config.example.json
borrrden: The file is still present

borrrden (changes requested):
Made progress, but needs a whole lot more detail in the spec.  Saying things like "start a replicator" is unhelpful.  What are the characteristics of the replicator?  See the other specs for an example.

spec/tests/QE/test_no_conflicts.md
borrrden: That's a key piece of information that's missing throughout this entire description.  There is nowhere that says the three databases are on three devices and I happen to know that this test you are porting has three databases on one device, hence the "simulate" phrasing.  If you are actually going to use multiple devices, then you are not "simulating" it, you are actually doing that and it should be reflected in the steps and the word "simulated" removed from the description.

spec/tests/QE/test_delta_sync.md
```diff
@@ -0,0 +1,164 @@
...
+Verify delta sync works with nested documents. Ensures that only changed nested documents are processed and replicated, and that delta sync bandwidth savings are realized.
+
+### Steps
+1. Reset SG and load `travel` dataset with delta sync enabled
+2. Reset local database, and load `travel` dataset.
+3. Start a replicator
+4. Wait until the replicator stops.
+5. Check that all docs are replicated correctly.
+6. Modify docs in CBL with nested docs
+7. Do push_pull replication
```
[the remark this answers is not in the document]
vipbhardwaj: Done, kindly resolve if LGTY.

spec/tests/QE/test_replicator_encryption_hook.md
vipbhardwaj: Oh? 
Even I do not know, the issue is I am relying on the documentation and more so on the descriptions and steps given to me by the testkit's steps. And those tests had this description for the test. So I assumed in their confidence that SGW cannot edit those fields. Even though now that I went back and checked, and I see that no-where do they actually try to edit and expect a failure for encrypted fields, not sure why did they wrote that then.

I have corrected that part of description, kindly resolve if LGTY.

spec/tests/QE/test_no_conflicts.md
vipbhardwaj: Done, kindly resolve if LGTY.

tests/QE/config.example.json
vipbhardwaj: My apologies, corrected now, kindly resolve if LGTY.

client/src/cbltest/api/syncgateway.py
```diff
@@ -870,3 +945,17 @@ async def close(self) -> None:
         """
         if not self.__admin_session.closed:
             await self.__admin_session.close()
+
+    async def create_public_session(
```
borrrden, line 949: This is adding to a bad design where the internal method `_send_request` is being used by tests.  If you need to use a different session, create a private one inside this class and expose API to do the things that you want.

tests/QE/test_delta_sync.py
```diff
@@ -0,0 +1,1247 @@
...
+        self, cblpytest: CBLPyTest, dataset_path: Path
+    ):
+        """
+        @summary: Verify push/pull replication works with large data
+            1. Create docs in CBL
+            2. Do push_pull replication
+            3. update docs in SGW  with/without attachment
+            4. Do push/pull replication
+            5. Verify number of docs updated using delta sync.
+        """
```
borrrden, line 36: It will be ok to not have a summary here.  It's just one more place where things can get stale.  This is the reason for the test spec and mark_test_step calls.

tests/QE/test_delta_sync.py
```diff
@@ -0,0 +1,1247 @@
...
+        updated_doc_size = len(json.dumps(sgw_doc_1.body).encode("utf-8")) + len(
+            json.dumps(sgw_doc_2.body).encode("utf-8")
+        )
+        delta_bytes_read = read_pull_bytes_after - read_pull_bytes_before
+        assert delta_bytes_read < updated_doc_size, (
+            f"Expected delta to be less than the full doc size, but got {delta_bytes_read} bytes (doc size: {updated_doc_size})"
+        )
+
+        await cblpytest.test_servers[0].cleanup()
+        self.mark_test_step("...COMPLETED...")
```
borrrden, line 222: No need for this

tests/QE/test_delta_sync.py
```diff
@@ -0,0 +1,1247 @@
...
+        updates = [
+            DocumentUpdateEntry(
+                "hotel_1",
+                None,
+                {"name": "SGW", "nested": {"name": "I am a nested field"}},
+            )
+        ]
+        await cblpytest.sync_gateways[0].upsert_documents(
+            "travel", updates, "travel", "hotels"
+        )
```
borrrden, line 330: This is going to be making redundant calls to sync gateway. You already have the revID information from getting the doc earlier, so why not use it directly rather than having upsert_documents ask for it again?

spec/tests/QE/test_delta_sync.md
```diff
@@ -0,0 +1,282 @@
...
+   * credentials: user1/pass
+8. Wait until the replicator stops.
+9. Record the bytes transferred
+10. Verify the new document is present in SGW
+11. Update docs in SGW:
+    * Update 2 airlines in `travel.airlines` with different text content
+    * Modify attachments in 2 airports in `travel.airports`
+12. Start the same replicator again.
+13. Wait until the replicator stops.
+14. Record the bytes transferred
```
borrrden, line 36: What is the double replication adding to this test?  Wouldn't it be enough to record the bytes transferred before and after step 7?

spec/tests/QE/test_delta_sync.md
```diff
@@ -0,0 +1,282 @@
...
+   * credentials: user1/pass
+8. Wait until the replicator stops.
+9. Record the bytes transferred
+10. Verify the nested document is present in SGW
+11. Update docs in SGW:
+    * Update different nested schedule fields in `travel.routes`
+12. Start the same replicator again:
+13. Wait until the replicator stops.
+14. Record the bytes transferred
+15. Verify delta sync bytes transferred is less than doc size.
```
borrrden, line 71: Same comment as previous test

tests/QE/test_no_conflicts.py
```diff
@@ -0,0 +1,569 @@
...
+        assert sg_doc is not None, "Document not found"
+        assert sg_doc.body.get("title") == "CBL Update 2", (
+            f"Wrong title in SG doc (expected 'CBL Update 2'; got {sg_doc.body.get('title')}"
+        )
+
+        await cblpytest.test_servers[0].cleanup()
+        self.mark_test_step("...COMPLETED...")
+
+    @pytest.mark.asyncio(loop_scope="session")
+    async def test_multiple_cbls_updates_concurrently_with_push(
```
borrrden, line 178: Careful with this, you only set up one test server in your topology and didn't mark this with min test servers so it is guaranteed to fail.

borrrden (changes requested):
I've made it through the first two test suites (delta_sync and no_conflicts)

tests/QE/test_no_conflicts.py
vipbhardwaj: So there are 3 tests in this class, 2 of which use 3 test servers, and the last one uses only 1 test server.
And I have marked this at the top of the class as : 
```
@pytest.mark.min_test_servers(3)
@pytest.mark.min_sync_gateways(1)
@pytest.mark.min_couchbase_servers(1)
```

Where I believe since it is the `min_test_servers` we are declaring for the class, so the test that uses 1 test server will also be functioning since there is no conflict with the max number of test servers.

As for the topology file, yes, since I have marked it, the test currently skips all 3 tests in this current class, and I am going to add more test servers once I know more devices are connected to the mac that hosts the Jenkins test for this.
After that I will add/update the topology with the device IDs and these tests will start running instead of being skipped.
For now it is to migrate the code. So that we remember adding the devices to the mac in US-office lab is the only thing remaining.

tests/QE/test_delta_sync.py
vipbhardwaj: This is not to ask for revID information. It is to create a small modification in the document.

As the document currently in SGW has : ```
doc = await cblpytest.sync_gateways[0].get_document(
            "travel", "hotel_1", "travel", "hotels"
        )
        assert doc is not None, "Document should exist in SGW"
        assert doc.body.get("name") == "CBL", "Document should have the correct name"
```
And the current `updates` has the following change in the `name` field : 
```
DocumentUpdateEntry(
                "hotel_1",
                None,
                {"name": "SGW", "nested": {"name": "I am a nested field"}},
            )
```

tests/QE/test_delta_sync.py
vipbhardwaj: Okay 👍🏻

tests/QE/test_delta_sync.py
vipbhardwaj: Okay 👍🏻 
Done, please mark resolved if LGTY.

spec/tests/QE/test_delta_sync.md
vipbhardwaj: 10 seconds worked as well and I have added the same to the test.
Please resolve if LGTY

tests/QE/test_delta_sync.py
vipbhardwaj: Done, resolving this conversation.

spec/tests/QE/test_delta_sync.md
vipbhardwaj: Answered in the previous comment.
https://github.com/couchbaselabs/couchbase-lite-tests/pull/189#discussion_r2131697778

tests/QE/test_delta_sync.py
borrrden: I think this is still kind of strange.  The second arg to DocumentUpdateEntry is the revID, but you are passing None when you already have received it above in `get_document` so couldn't you just make an update directly by passing `doc.revid` to the update entry and then using `update_documents`?  Upsert documents seems to go and retrieve all the docs it wants to update first, which you have already done above with `get_document`.

spec/tests/QE/test_delta_sync.md
borrrden: You already have some existing documents so wouldn't it be faster to simply update one of them and measure the delta during the push?  Rather than creating, pushing, modifying, and then pulling?  Or if you are only interested in pull for some reason update one of them on the sgw side and measure the delta during the pull.  It seems like this adds an unneeded replication to the process which bit by bits adds to total test runtime.

tests/QE/test_no_conflicts.py
borrrden: Oh I missed it at the class level.  I always get a bit nervous that it won't work if it's declared there but if you've observed it working I believe it.

spec/tests/QE/test_delta_sync.md
vipbhardwaj: Took me a while but yes you're right, that approach achieves the same goal with one less replication and redundant checks per test. I have optimized the tests accordingly. Thanks for the explanation though, it really helped.

spec/tests/QE/test_delta_sync.md
vipbhardwaj: Fixed the `double replication` inefficiency in the tests. Using first and second PULL to compute the delta for an-existing document change. Ran the tests and checked that it achieves the same goals, only faster.

borrrden (approved):
Assuming that these tests reliably pass I think this is good to go, but I don't like the sgw throwing an exception for a 404.  It should just return null and it is probably a bug that it does not.  I'll try to address that later.

PR #190, opened 2025-05-19
Make mypy tests pass

torcolvin:
- remove client/requirements.txt since it is described in client/pyproject.toml
- add jenkins specific dependencies
- update mypy github file to make sure dependencies are specified correctly in each environment
- switch to uv because it is faster

Fixes the issue where mypy tests won't run on https://github.com/couchbaselabs/couchbase-lite-tests/pull/189 . It is definitely valid to have conftest.py in multiple directories and not worth linting over it.


vipbhardwaj:
Can we keep `--check-untyped-defs` skipped in the new draft at line 26 in the bash file?
Was that flexibility in checks intentional?

vipbhardwaj (approved):
LGTM

torcolvin:
> Can we keep `--check-untyped-defs` skipped in the new draft @ verify_python.sh:27 ? Was that flexibility in checks intentional?

Adding: `check_untyped_defs = true` to pyproject.toml root. This is a more restrictive check, wher

PR #195, opened 2025-06-04
Write 1.2.1 of the REST API spec

borrrden:
This basically adds the ability to set a CA certificate authenticator and identity in the call to startMultipeerReplciator.  The TDK client will provide the cert and private key to use for both, so the test server need only do the following:

1. Convert the PEM bytes of the certificate into an appropriate platform type for the authenticator
2. Convert Base64 PKCS#12 data into bytes for using in TLSIdentity.CreateIdentity

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -846,7 +846,7 @@ paths:
           application/json:
             schema:
               type: object
-              required: ['peerGroupID', 'database', 'collections']
+              required: ['peerGroupID', 'database', 'collections', 'identity']
               properties:
                 peerGroupID:
                   type: string
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -46,6 +46,9 @@ info:
       ```
 
     Changes
+    1.2.1 (06/04/2025)
+    * Add authenticator property to [/startMultipeerReplicator](#operation/startMultipeerReplicator)
+
     1.2.0 (04/09/2025)
     * Add [/startMultipeerReplicator](#operation/startMultipeerReplicator),
     * Add [/stopMultipeerReplicator](#operation/stopMultipeerReplicator) endpoint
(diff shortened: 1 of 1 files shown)


spec/api/api.yaml
```diff
@@ -833,11 +836,12 @@ paths:
...
       summary: Starts a Multipeer replicator
       description: |-
-        Starts a P2P listener so that the device can act as the passive side of a P2P replication.
+        Starts a multipeer replicator to create a mesh of connected clients.
       operationId: startMultipeerReplicator
       requestBody:
         description: |-
-          The request object containing the collections to share and, optionally, the port to listen on.
+          The request object containing the collections to share and the peer group ID.  Optionally,
+          it can contain an authenticator as well.  The default (null) behavior is accept all connections.
```
pasin, line 844: Is identity optional or mandatory?

borrrden: Optional. It’s not in the list of required fields.

spec/api/api.yaml
```diff
@@ -1276,6 +1284,38 @@ components:
...
+    MultipeerReplicatorIdentity:
+      description: |-
+        MultipeerReplicatorIdentity is used to specify a TLS certificate and key particular
+        for a multipeer replciator so that it can identify itself.
+      type: object
+      required: ['encoding', 'data']
+      properties:
+        encoding:
+          type: string
+          enum: ['PKCS12']
```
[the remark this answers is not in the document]
borrrden: Yes I think all platforms can handle this but i wanted to send over the type of encoding just in case.

spec/api/api.yaml
borrrden: Made it mandatory but yes a new self signed one will be generated every time, just on the TDK client side now

PR #199, opened 2025-06-07
Remove the dataset parameter from the prebuild logic

borrrden:
This means taking the dataset version out of the filename and not having a Jenkins prebuild option to specify it anymore.  Prebuilt test servers will hard code 3.2 for now until the copy dataset logic can be fully removed.


borrrden:
This looks bigger than it actually is because a lot of stuff needs to be repeated across the different test server platforms.  Basically, removing an argument from the abstract base constructor which is then echoed into the subclasses.  Also removing the dataset version from copy_datasets.  copy_datasets is now deprecated and will be removed once all platforms have asset downloading in place.

bmeike (approved):
This is because the dataset is now chosen at runtime?

borrrden:
> This is because the dataset is now chosen at runtime?

Precisely.  The current prebuild system is designed to make two builds (one for 3.2 and one for 4.0), and I simplified it to make only one (dataset chosen at runtime).

PR #200, opened 2025-06-11
.NET prebuild logic and quality of life improvements

borrrden:
For all platforms this fixes an issue whereby archiving artifacts on multiple platforms all overwrite each other.  It also overhauls the use of create_env / New-Venv to automatically download the needed python, as well as install uv for faster pip install speed.  

For .NET a bug was fixed whereby locking was crossing async boundaries and causing exceptions because of a new thread finishing the async job and trying to release a lock it didn't own.

Fix a prebuilding bug where the keychain was locked for builds making an iOS artifact.


jenkins/pipelines/dev_e2e/c/teardown.ps1
```diff
@@ -1,8 +1,21 @@
...
+
+Stop-Venv
+New-Venv venv
 .\venv\Scripts\activate
-pip install -r requirements.txt
+uv pip install -r requirements.txt
 python .\stop_backend.py --topology topology_setup\topology.json
-Pop-Location
\ No newline at end of file
+Cleanup
```
pasin, line 21: Does it still need to call Cleanup here along with the `trap`?

pasin (approved):

bmeike (approved):
:shrug:  If there is anything wrong with this, I'm not gonna notice it.

PR #204, opened 2025-06-11
Convert Java web service to a JAR

borrrden:
So that it can be prebuilt


servers/jak/webservice/app/src/main/java/com/couchbase/lite/javaws/mobiletest/EmbeddedJettyServer.java
```diff
@@ -0,0 +1,22 @@
+package com.couchbase.lite.javaws.mobiletest;
```
bmeike, line 1: I do not think this should be a new class/file.  I think that the `main` method should just be added to `TestServerApp.java` similar to the way the main method is set up in the desktop version

bmeike (changes requested):
I think it will work to put that main method into the TestServerApp class.

servers/jak/webservice/app/src/main/java/com/couchbase/lite/javaws/mobiletest/EmbeddedJettyServer.java
borrrden: Done!

bmeike (approved):
Yep!

PR #214, opened 2025-06-16
First draft for system test - CBL migration

vipbhardwaj:
Even though we know TDK - logslurp cannot support such a test right now, it is okay to have it merged and : 

- Run it without LogSlurp (everything else runs fine, I tested a 6 hour scenario locally ✅ )
- Or, if some extreme reason to block, then not run it till the reason is mitigated.

### DIFF
--- a/tests/QE/test_system.py
+++ b/tests/QE/test_system.py
@@ -1,0 +1,597 @@
+import asyncio
+import random
+import threading
+from datetime import datetime, timedelta
+from pathlib import Path
+
+import pytest
+from cbltest import CBLPyTest
+from cbltest.api.cbltestclass import CBLTestClass
+from cbltest.api.cloud import CouchbaseCloud
+from cbltest.api.database_types import DocumentEntry
+from cbltest.api.replicator import Replicator
+from cbltest.api.replicator_types import (
+    ReplicatorActivityLevel,
+    ReplicatorBasicAuthenticator,
+    ReplicatorCollectionEntry,
+)
+
+
+@pytest.mark.min_test_servers(1)
+@pytest.mark.min_sync_gateways(1)
+@pytest.mark.min_couchbase_servers(1)
+class TestSystem(CBLTestClass):
+    # Test parameters
+    NUM_OF_DOCS = 1000000  # 1M docs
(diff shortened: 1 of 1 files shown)


tests/QE/test_system.py
```diff
@@ -0,0 +1,597 @@
...
+            f"[{db_name}] Creating {docs_per_db} documents in {iterations_per_db} batches"
+        )
+
+        for iteration in range(iterations_per_db):
+            max_retries = 3
+            retry_delay = 1
+            batch_doc_ids: list[str] = []
+
+            # Retry logic with exponential backoff - wraps the entire batch operation
+            for retry in range(max_retries):
```
borrrden, line 55: What is the expected way that this is going to fail?

tests/QE/test_system.py
```diff
@@ -0,0 +1,597 @@
...
+                ReplicatorActivityLevel.IDLE,
+                timedelta(seconds=self.REPL_STATUS_CHECK_SLEEP_TIME),
+                timedelta(minutes=3),  # Shorter timeout for individual threads
+            )
+            if status.error:
+                print(
+                    f"[{db_name}] Warning: Replication error during wait: {status.error}"
+                )
+        except Exception as e:
+            print(f"[{db_name}] Warning: Error waiting for replication idle: {e}")
```
borrrden, line 579: What is the point of doing this?  This will make this error harder to find since it will only be present in the jenkins console, and when the test fails it will fail for a different reason.

borrrden (changes requested):
Needs a spec in the specs folder to start with

vipbhardwaj:
@borrrden We should honestly leave this PR since it is almost a year old as far as I remember. Hence it lacks everything that's enhanced and added to TDK, inlcuding, rosmar. Plus the fact that this is focused on CBL system tests which are already there by Barkha makes this code further redundant.

If we do need this, I'd actually trash this PR anyways and write a system test from scratch.
There's a discussion today about SGW testing, which is my changed domain now, there it'll be decided to go with a SGW system test or not. Hence this PR is from Bo-chun's time and I'm embaressed at my own code and its just a very stupid PR.

PR #222, opened 2025-06-20
API 2.0 spec

bmeike:
Replace dataset_version with dataset and blob urls

### DIFF
--- a/spec/api/api.yaml
+++ b/spec/api/api.yaml
@@ -46,6 +46,11 @@ info:
       ```
 
     Changes
+    2.0.0 (06/13/2025)
+    * This is a breaking change: CBLTest-API-Version is now "2"
+    * deprecate dataset_version
+    * In ResetConfiguration, change the value for the "dataset" key to a url.
+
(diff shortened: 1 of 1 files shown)


spec/api/api.yaml
```diff
@@ -1044,7 +1050,7 @@ components:
             example: ['people[22].address', 'people[3]']
         updatedBlobs:
           description: A dictionary whose keys are the key paths into the document to be updated with the
-            blob objects and values are the blob keys in the dataset.
+            blob objects and values are the urls from which to obtain the blob files.
           type: object
           additionalProperties: {}
           example: {'photo.thumbnail': 's1.jpg'}
```
torcolvin, line 1056: is this correct if this is a url? `s1.jpg` doesn't look like a url to me.

bmeike: Yes.  I missed one.  Will fix.

bmeike: fixed

torcolvin (approved):

PR #227, opened 2025-06-25
Remove final parts of the dataset version parameter

borrrden:
This is the same thing repeated across 12 different folders.  Removing dataset_version and its variants from the Jenkinsfile, topology, run tests script(s) and setup_test.py.  Reviewers can look at their own implementation and need not worry about others.


borrrden:
Not much I can do about that.  It's the same 6-7 files, but a dozen or so variations of them across dev and QE

PR #228, opened 2025-07-01
Hardcode the 3.2 dataset value until it is removed

borrrden:
It's still used at the beginning of a test run to choose which versions of datasets to download.


pasin (approved):

vipbhardwaj (approved):
LGTM 👍🏻

PR #237, opened 2025-07-30
Add quick start script to run local docker

torcolvin:
The intention behind this quick start script is to be able to start an environment quickly and run a test. This is a bit of a work in progress and I use it personally to run test code.

This will:

1. build testserver (currently only C supported)
2. start testserver in the background
3. start docker containers
4. create a `config.json` locally to be able to use everything

Then you can do a single command to run tests: `uv run pytest ...`

```
#   uv run -- quick_start_local.py --test-server-type c --cbl-version 3.2.4 --cbl-build-number 9 --dataset-version 3.2
#   uv run pytest --config config.json tests/dev_e2e/test_basic_replication.py
```

TODO:

[ ] Something seems a bit wrong with launching test server in the background, I did a quick hack in python and it sort of works but sometimes testserver.log is polluted with pytest output.

### DIFF
--- a/environment/docker-compose.yml
+++ b/environment/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     build:
       context: ./cbs
       args:
-        - COUCHBASE_VERSION=${COUCHBASE_VERSION:-7.6.4}
+        - COUCHBASE_VERSION=${COUCHBASE_VERSION:-7.6.7}
     ports:
       - "8091:8091"
       - "8092:8092"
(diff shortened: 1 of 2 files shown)


quick_start_local.py
```diff
@@ -0,0 +1,120 @@
...
+        ],
+        "sync-gateways": [{"hostname": "localhost", "tls": False}],
+        "couchbase-servers": [{"hostname": "localhost"}],
+        "api-version": 1,
+    }
+    with pathlib.Path("config.json").open("w") as f:
+        json.dump(config, f, indent=2)
+
+
+def start_testserver(server_type: TestServerType) -> None:
```
borrrden, line 78: FYI I have put extensive work into having this functionality available in python so you might consider importing and using that instead of rewriting it: 

See the environment/aws/topology_setup folder (note that despite being in the AWS folder, this particular area has no AWS dependencies)

PR #245, opened 2025-09-05
CBL-7166 : Implement Merge-Dict conflict resolver in C TestServer

pasin:
* Implemented MergeDictConflictResolver.

* Made variable names more clear in the existing MergeConflictResolver.

* Corrected the Merge_Dict spec for error cases.


borrrden (changes requested):
I don't see any form of error checking for conflicting values

pasin:
True, I totally forgot about that.

PR #252, opened 2025-09-25
Add a method to create a replication URL through load balancer

borrrden:


client/src/cbltest/api/syncgateway.py
```diff
@@ -419,14 +419,18 @@ def tls_cert(self) -> str | None:
...
 
         :param db_name: The DB to replicate with
         """
         _assert_not_null(db_name, "db_name")
-        return urljoin(self.__replication_url, db_name)
+        sgw_address = urljoin(self.__replication_url, db_name)
+        if not load_balancer:
+            return sgw_address
+
+        return sgw_address.replace("wss", "ws").replace(self.__hostname, load_balancer)
```
pasin, line 433: I think I don't understand how the TLS handshake works when load balancer is used as the URL will be switched to ws://.

pasin (approved):

PR #263, opened 2025-10-06
CBL-7340 : E2E test timeout update as replication taking time

Aniket392:
- Updated the max tries and wait time
- Updated the case of no collection passed, use default collection
- Updating the Logger.get which can be nullable
- Changed Dataset version from 3.2 to 4.0


borrrden:
Is this still being worked on?

PR #287, opened 2025-10-31
Edge server tests, provisioning scripts and APIs

sanjivanipatrax:
This PR adds test coverage and automation support for Edge Server, including new test cases, helpers, provisioning logic, and installation scripts for end-to-end validation.

**Features added:**
- End-to-end Edge Server test coverage across authentication, CRUD, replication, conflict resolution, and system verification scenarios (in `tests/QE/edge_server/`).
- Helper modules for REST API interactions, configuration generation, and remote shell operations (in `client/src/cbltest/api/` and `client/src/cbltest/`).
- Environment setup scripts to install, uninstall, and configure Couchbase Server, Sync Gateway, and Edge Server components (in `environment/cbs/`, `environment/sg/`, and `environment/edge_server/`).
- VM provisioning from the Mobile QE VM pool (in `environment/vm_management/`).


client/src/cbltest/api/remoteshell.py
```diff
@@ -0,0 +1,183 @@
+import subprocess
```
borrrden: I'm very skeptical about having tests that actually connect to and execute commands on a remote machine because now there are assumptions here about the environment on the other side.  I guess there might be no other way to do this for edge server but at least I'd like to keep this as minimal as possible.  Having a class called "remote shell" is vague and hard to understand.

environment/cbs/uninstall_couchbase.py
```diff
@@ -0,0 +1,139 @@
+import json
```
borrrden: This is also redoing a bunch of orchestrator work, as well as doing some weird stuff with /etc/hosts....

environment/vm_management/vm_pool_manager.py
```diff
@@ -0,0 +1,366 @@
+import logging
```
borrrden: Is there any benefit to using the QE VM pool rather than AWS?

resources/images/20mb.jpg
borrrden: This definitely needs to be set up to use Git LFS, it's gigantic...

client/src/cbltest/api/httpclient.py
```diff
@@ -0,0 +1,694 @@
+import asyncio
```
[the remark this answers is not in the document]
sanjivanipatrax: For edge server testing, we are using HTTP clients instead of test servers. This class helps us create several clients per VM (about 10–20) and make them send requests at the same time to different components.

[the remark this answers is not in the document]
borrrden: I don't see how that will create any sort of reliable testing.  Is that any different than using 10-20 threads inside of test server to send requests?  I worry about fragmentation and opaque hard to understand things like "http client" which doesn't really describe its purpose.

PR #383, opened 2026-04-10
fix: conditional /_config call to not mess with SyncGatewayUserClient

vipbhardwaj:
[CBG-5281](https://jira.issues.couchbase.com/browse/CBG-5281)

the rosmar call to _config was added in the past, but it was at the `_SyncGatewayBase` class level causing even the `SyncGatewayUserClient` to end up calling it which led tests to fail since the PUBLIC PORT cannot call that endpoint but the port assigned to User Client subclass uses public port itself. 
I have added a condition so that if the port is admin then only shall the call go to config check for rosmar bool value. Since otherwise it would fail rosmar tests too regardless. This is backwards compatible as well since if the port is PUBLIC coming from the `SyncGatewayUserClient`, there is no point in calling `/_config`

[CBG-5281]: https://couchbasecloud.atlassian.net/browse/CBG-5281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -175,6 +175,16 @@ class DocumentRevision:
         """
         Compares two revisions, automatically using the correct mathematical logic
         based on the detected format.
+
+        For SGW 3.x: Compares generation numbers (higher generation = newer).
+        For SGW 4.x: Only meaningful comparison is when both vectors originate
+        from the same source ID. Within the same source, timestamps are monotonically
+        increasing. Cross-source comparisons are not meaningful since:
+        - Each src has its own logical clock (not synced with others)
+        - SrcIDs are MD5 hashes per device/db with no meaningful ordering
+
+        :raises ValueError: If comparing version vectors from different sources or
+                          mixed revision formats
         """
         if self.is_version_vector != other.is_version_vector:
             raise ValueError(
@@ -182,12 +192,17 @@ class DocumentRevision:
             )
 
         if self.is_version_vector:
-            # SGW 4.0 LWW Logic: Compare timestamps, fallback to source ID
-            if self.timestamp_int == other.timestamp_int:
-                return self.source_id > other.source_id
+            # SGW 4.0 Version Vector comparison
+            # Only meaningful when both vectors are from the same source ID
+            if self.source_id != other.source_id:
+                raise ValueError(
+                    f"Cannot compare version vectors from different sources: "
+                    f"{self.raw_value} (source={self.source_id}) vs "
+                    f"{other.raw_value} (source={other.source_id}). "
+                    f"Each source has its own logical clock."
+                )
             return self.timestamp_int > other.timestamp_int
         else:
-            # SGW 3.x Logic: Compare generation integers
             return self.generation > other.generation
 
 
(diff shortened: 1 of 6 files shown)


client/src/cbltest/api/syncgateway.py
```diff
@@ -556,20 +600,25 @@ def __init__(
...
+            r.raise_for_status()
+            try:
+                self.using_rosmar = r.json()["bootstrap"]["server"].startswith("rosmar")
+            except AttributeError:
+                raise CblTestError(
+                    "Unexpected response {r.json()} from Sync Gateway /_config endpoint, cannot determine if using Rosmar"
+                ) from None
+        else:
+            # For public port clients, default to not using rosmar
             self.using_rosmar = False
```
torcolvin: If this is the requirement then this check for `using_rosmar` needs to be moved to the admin only section of the SyncGateway class, because it would not be correct to saying `using_rosmar` is False anytime. The `SyncGatewayUserClient` shouldn't need to know if rosmar is being used or not, this is required for the purposes of deciding how to instantiate CouchbaseCloud and will be needed for some calls around how to destroy buckets. https://github.com/couchbase/sync_gateway/pull/8158

client/src/cbltest/api/syncgateway.py
```diff
@@ -147,6 +147,50 @@ def to_json(self) -> Any:
         return self.__config
 
 
+class DocumentRevision:
```
torcolvin, line 150: I don't know if this code belongs in this PR and I'd be curious to see how it is used because > is not necessarily a meaningful operation.

This isn't a DocumentRevision so much as a RevisionID. I can see why you would want to compare them, but I don't think that a greater than comparison is valid generally.

Additionally, these are very much not specific to Sync Gateway and they are used by all of mobile. I do not believe couchbase lite generally will expose these through a public API, but it is definitely possible edge server or couchbase lite js do.

client/src/cbltest/api/syncgateway.py
```diff
@@ -147,6 +147,50 @@ def to_json(self) -> Any:
...
+        if self.is_version_vector != other.is_version_vector:
+            raise ValueError(
+                f"Cannot compare mixed revision formats: {self.raw_value} vs {other.raw_value}"
+            )
+
+        if self.is_version_vector:
+            # SGW 4.0 LWW Logic: Compare timestamps, fallback to source ID
+            if self.timestamp_int == other.timestamp_int:
+                return self.source_id > other.source_id
+            return self.timestamp_int > other.timestamp_int
```
torcolvin: You can only meaningfully do a greater than if the source id is the same.

For each source id, the clocks are monotonic, but it is not guaranteed at all that the source ids will have matching clocks.

Additionally, depending on where you are getting the timestamp component from, these have very different formats for storage, this could be a plan integer, this could be little endian hex. This matters a lot for the caller.

source IDs are md5 hashes per device (for SG) and per database (CBL) and don't have remotely meaningful > comparisons.

vipbhardwaj: Oh I see, I will correct that then.

client/src/cbltest/api/syncgateway.py
vipbhardwaj: I see, although I was working on both the things at the same time, different halves of the day. Therefore they both got pushed and yes this is not a part of the issue at hand right now.

The reason I added it is because I had to import the `Version` class and then fetch SGW version and compare them in every test to validate both version vectors and rev_ids based on the SGW version being test. Its needed to maintain backwards compatibility since the tests have to be in the middle of 3.x & 4.x for many months atleast for now.

client/src/cbltest/api/syncgateway.py
vipbhardwaj: Alright! will do.

client/src/cbltest/api/syncgateway.py
vipbhardwaj: I am keeping it for now since most of the tests do rely on either changing once or a load of concurrent updates on CBL or SGW sides, then verifying `_changes` and the `rev` to be higher, amongst other param checks.
I had to manually write repetitive code hence I made this class to de-clutter tests overall.

torcolvin: Can this be split out from the rosmar conditional PR? It's really helpful for reviewing to only have a PR do one thing and be small.

I think it's good to have a standard function to compare versions, I just want to not have a `is_newer_than` function.

vipbhardwaj: That makes sense, I've been getting closer the habit of "smallest-PRs" , but there's some way to go for me.
Anyways, so I make your PR as the parent of this one. But do I also remove my code changes since they're in yours too?

vipbhardwaj:
Hey, @torcolvin since the JIRA to this is also now shifted to resolved thanks to your PR, is it better if we just drop this PR now?
I tested it and am not liking the whole DocumentVersion class stuff as well, it'd need more refining before I can use it properly considering I just started it as a side thing to target stuff better, and I have bigger things coming up now.

PR #384, opened 2026-04-13
Add URLEndpointListener QE tests and TLS support in iOS TestServer

barkha06:
- Added QE test coverage for URLEndpointListener, migrated from mobile-testkit to validate legacy peer-to-peer functionality
- Implemented TLS support in iOS TestServer for URLEndpointListener
- Enabled execution of peer-to-peer replication tests with TLS
- Added corresponding spec files

### DIFF
--- a/client/src/cbltest/api/listener.py
+++ b/client/src/cbltest/api/listener.py
@@ -41,9 +41,9 @@ class Listener:
 
         self.__original_port = port
         self.__identity = identity
+        self.__index = database._index
         if not disable_tls and not identity:
             self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
-        self.__index = database._index
         self.reuse_identity = reuse_identity
         self.__request_factory = database._request_factory
         self.__tracer = get_tracer(__name__, VERSION)
--- a/client/src/cbltest/api/listener.py
+++ b/client/src/cbltest/api/listener.py
@@ -19,8 +19,7 @@ class Listener:
         collections: list[str],
         port: int = 59840,
         disable_tls: bool = False,
-        identity: CertKeyPair | None = None,
-        reuse_identity: bool = False,
+        identity: CertKeyPair | None = None
     ):
         self.database = database
         """The database that the listener will be serving"""
@@ -43,9 +42,6 @@ class Listener:
         self.__original_port = port
         self.__identity = identity
         self.__index = database._index
-        if not identity:
-            self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
-        self.reuse_identity = reuse_identity
         self.__request_factory = database._request_factory
         self.__tracer = get_tracer(__name__, VERSION)
         self.__id: str = ""
@@ -53,7 +49,8 @@ class Listener:
     @property
     def identity(self) -> CertKeyPair:
         """Gets the identity used by the replicator"""
-        assert self.__identity is not None, "Listener identity not initialized"
+        if self.__identity is None:
+            self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
         return self.__identity
 
     async def start(self) -> None:
@@ -65,8 +62,7 @@ class Listener:
                 collections=self.collections,
                 port=self.port,
                 disable_tls=self.disable_tls,
-                identity=self.__identity,
-                reuse_identity=self.reuse_identity,
+                identity=self.__identity
             )
             resp = await self.__request_factory.send_request(self.__index, request)
             if resp.error is not None:
--- a/client/src/cbltest/api/listener.py
+++ b/client/src/cbltest/api/listener.py
@@ -17,7 +17,7 @@ class Listener:
         self,
         database: Database,
         collections: list[str],
-        port: int = 59840,
+        port: int | None = None,
         disable_tls: bool = False,
         identity: CertKeyPair | None = None,
     ):
@@ -27,7 +27,7 @@ class Listener:
         self.collections = collections
         """The collections within the database that the listener will be serving"""
 
-        self.port = port
+        self.__port = port
         """
         The port that the listener will request to listen on
         (if None, the OS will choose).  Once start is called,
@@ -52,6 +52,11 @@ class Listener:
         assert self.__identity is not None, "Listener identity not initialized"
         return self.__identity
 
+    @property
+    def port(self) -> int:
+        assert self.__port is not None, "Listener port not set"
+        return self.__port
+
     def set_identity(self):
         self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
 
@@ -62,7 +67,7 @@ class Listener:
                 TestServerRequestType.START_LISTENER,
                 db=self.database.name,
                 collections=self.collections,
-                port=self.port,
+                port=self.__port,
                 disable_tls=self.disable_tls,
                 identity=self.__identity,
             )
@@ -73,7 +78,7 @@ class Listener:
                 return
 
             cast_resp = cast(PostStartListenerResponseMethods, resp)
-            self.port = cast_resp.port
+            self.__port = cast_resp.port
             self.__id = cast_resp.listener_id
 
     async def stop(self) -> None:
@@ -89,5 +94,5 @@ class Listener:
                 cbl_trace(resp.error.message)
                 return
 
-            self.port = self.__original_port
+            self.__port = self.__original_port
             self.__id = ""
--- a/client/src/cbltest/api/listener.py
+++ b/client/src/cbltest/api/listener.py
@@ -3,11 +3,12 @@ from typing import cast
 from opentelemetry.trace import get_tracer
 
 from cbltest.api.database import Database
+from cbltest.api.x509_certificate import CertKeyPair, create_leaf_certificate
 from cbltest.logging import cbl_error, cbl_trace
 from cbltest.requests import TestServerRequestType
 from cbltest.response_types import PostStartListenerResponseMethods
 from cbltest.version import VERSION
-from cbltest.api.x509_certificate import CertKeyPair, create_leaf_certificate
+
 
 class Listener:
     """A class representing the passive side of a replication inside of a test server"""
@@ -64,7 +65,7 @@ class Listener:
                 port=self.port,
                 disable_tls=self.disable_tls,
                 identity=self.__identity,
-                reuse_identity=self.reuse_identity
+                reuse_identity=self.reuse_identity,
             )
             resp = await self.__request_factory.send_request(self.__index, request)
             if resp.error is not None:
--- a/client/src/cbltest/api/listener.py
+++ b/client/src/cbltest/api/listener.py
@@ -19,7 +19,7 @@ class Listener:
         collections: list[str],
         port: int = 59840,
         disable_tls: bool = False,
-        identity: CertKeyPair | None = None
+        identity: CertKeyPair | None = None,
     ):
         self.database = database
         """The database that the listener will be serving"""
@@ -50,8 +50,11 @@ class Listener:
     def identity(self) -> CertKeyPair:
         """Gets the identity used by the replicator"""
         if self.__identity is None:
-            self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
+            self.set_identity()
         return self.__identity
+
+    def set_identity(self):
+        self.__identity = create_leaf_certificate(f"Test Server {self.__index}")
 
     async def start(self) -> None:
         """Start listening for incoming connections"""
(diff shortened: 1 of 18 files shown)


servers/ios/TestServer/Utils/DatabaseManager.swift
```diff
@@ -107,7 +107,36 @@ class DatabaseManager {
         var listenerConfig = URLEndpointListenerConfiguration(collections: collectionsArr)
         listenerConfig.port = port
         listenerConfig.disableTLS = disableTLS
-        
+        if !listenerConfig.disableTLS {
+        let label = "ios-p2p-\(dbName)"
+        guard let data = Data(base64Encoded: identity.data) else {
+            throw TestServerError.badRequest("Invalid replicator identity data")
```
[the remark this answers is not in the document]
barkha06: The automation always generates an identity despite TLS being disabled or enabled.

client/src/cbltest/api/listener.py
```diff
@@ -16,8 +17,10 @@ def __init__(
         self,
         database: Database,
         collections: list[str],
-        port: int | None = None,
+        port: int = 59840,
```
borrrden: I don't see a reason to make this have a default.

client/src/cbltest/api/listener.py
```diff
@@ -16,8 +17,10 @@ def __init__(
...
         disable_tls: bool = False,
+        identity: CertKeyPair | None = None,
+        reuse_identity: bool = False,
```
borrrden: This property doesn't really sit well with me.  It's awkward.  I can't think of any situation in which you'd want to forcibly recreate an identity, so why not just have it send null when you want to reuse (assuming disableTLS is false).

servers/ios/TestServer/ContentTypes/StartListenerRequest.swift
```diff
@@ -13,13 +13,16 @@ extension ContentTypes {
         let collections: [String]
         let port: UInt16?
         let disableTLS: Bool?
+        let identity: MultipeerReplicatorIdentity?
```
pasin: If this is shared with the URLEndpointListener, we should rename this to something like `TLSIdentityData` (avoid using TLSIdentity as it will be conflicted with the CBL-Swift's TLSIdentity.

servers/ios/TestServer/ContentTypes/StartListenerRequest.swift
```diff
@@ -13,13 +13,16 @@ extension ContentTypes {
...
+        let reuseIdentity: Bool?
```
pasin: I don't feel like having an ability to reuse the identity is a good idea as tests should be independent with each other.

servers/ios/TestServer/Handlers/StartListener.swift
```diff
@@ -17,7 +17,8 @@ extension Handlers {
         
         let dbManager = req.databaseManager
         let disableTLS = listenerStartRq.disableTLS ?? false
-        let id = try dbManager.startListener(dbName: listenerStartRq.database, collections: listenerStartRq.collections, port: listenerStartRq.port, disableTLS: disableTLS)
+        let reuseIdentity = listenerStartRq.reuseIdentity ?? false
+        let id = try dbManager.startListener(dbName: listenerStartRq.database, collections: listenerStartRq.collections, port: listenerStartRq.port, disableTLS: disableTLS,identity: listenerStartRq.identity,reuseIdentity: reuseIdentity)
```
pasin: This line is very long, please format it into multiple lines.

servers/ios/TestServer/Utils/DatabaseManager.swift
```diff
@@ -107,7 +107,36 @@ class DatabaseManager {
...
+            } else {
+                importedIdentity = try TLSIdentity.importIdentity(
+                    withData: data,
+                    password: identity.password,
+                    label: label
+                )
+            }
+        } catch {
+            throw TestServerError.badRequest("Failed to import TLS identity")
+        }
```
pasin, line 130: This code doesn't have proper indentation. Please fix.

Also there is `private static func multipeerReplicatorIdentity(for config: ContentTypes.MultipeerReplicatorConfiguration) throws -> TLSIdentity` function that you can reuse for this purpose. But I would suggest to change the function to `createTLSIdentity(for identityData: TLSIdentityData label: String) throws -> TLSIdentity`.

pasin (changes requested):
I have review the iOS TestServer code. Please see my comment.

borrrden (changes requested):

client/src/cbltest/api/listener.py
barkha06: This is a required parameter because the replication URL needs to be generated based on the port, so the ty lint check failed since it was marked optional but it needs to have int value later.

client/src/cbltest/api/listener.py
barkha06: This property is needed for one of the tests which should stop and restart mid p2p. And since the general logic is to delete an existing identity, to bypass it we need to send a parameter explicitly. As for why not Null, the ty lint check refused to have it as null because I have set the identity as a property. This was done so that when setting replication config, the pinned server cert comes from the identity.

servers/ios/TestServer/ContentTypes/StartListenerRequest.swift
barkha06: The reuse Identity parameter is used in only one test and doesn't rely on previous tests. It's needed for one of the tests that stops a listener and starts it again. Since the replicator already has a pinned cert, unless the listener starts with the same identity, the test fails due to authentication error and by design of the test a new replicator is not to be defined.

servers/ios/TestServer/Handlers/StartListener.swift
barkha06: Yes sure, will do

servers/ios/TestServer/Utils/DatabaseManager.swift
barkha06: Okay sure I'll make this change

client/src/cbltest/api/listener.py
borrrden: It will still have an int value after it is started.  I think the ty lint isn't getting all of the information.

https://github.com/couchbaselabs/couchbase-lite-tests/blob/d19e200e8ad034752cfeacb03a2a44d561f6ed38/client/src/cbltest/api/listener.py#L78

client/src/cbltest/api/listener.py
borrrden: I don't really follow why are are bending over for a linter.  There should be some way to satisfy it without resorting to the weird situation in which a parameter causes a previous one to be ignored arbitrarily.  identity in the above shows as optional, so if TLS is not disabled the natural thing to do if the identity is null would be to check if it already has an identity and use the one it finds, and have it be an error if no identity is found.  This means not auto generating an identity inside the constructor, and instead creating it in the test but the result is a less awkward API.

borrrden:
Also, regardless since you've changed the REST API you need to modify the OpenAPI spec and bump the version for the iOS test server and TDK client.

client/src/cbltest/api/listener.py
barkha06: I have modified the code to set the identity in tests and removed reuse_identity

servers/ios/TestServer/Utils/DatabaseManager.swift
```diff
@@ -86,7 +86,7 @@ class DatabaseManager {
         }
     }
     
-    public func startListener(dbName: String, collections: [String], port: UInt16?, disableTLS: Bool = false) throws -> UUID {
+    public func startListener(dbName: String, collections: [String], port: UInt16?, disableTLS: Bool = false, identity:ContentTypes.TLSIdentityData?) throws -> UUID {
```
pasin: Formatting : Can you add a space after `identity: `?

servers/ios/TestServer/Utils/DatabaseManager.swift
```diff
@@ -107,7 +107,27 @@ class DatabaseManager {
...
-        
+        if !listenerConfig.disableTLS {
+            let label = "ios-p2p-\(dbName)"
+            let importedIdentity: TLSIdentity
+            
+            do {
+                if let id = identity {
+                    importedIdentity = try DatabaseManager.createTLSIdentity( for: id, label:label)
+                } else {
+                    guard let existingIdentity = try TLSIdentity.identity(withLabel: label) else {
```
pasin, line 118: Let do not do this logic to avoid using any stale identity which may cause a test failure that is difficult to figure out. So if the created identity is nil, we can throw the error right away.

pasin: You can do something like this 

```Swift
listenerConfig.disableTLS = disableTLS

if let id = identity {
    guard let tlsIdentity = try DatabaseManager.createTLSIdentity( for: id, label:label) else { 
        throw TestServerError.badRequest("Failed to import TLS identity")
    }
    listenerConfig.tlsIdentity = tlsIdentity
}

```

client/src/cbltest/api/listener.py
borrrden: I still don't get why the above is not enough to satisfy this situation.

pasin (changes requested):

spec/api/api.yaml
borrrden, line 1: This needs to have an entry in the changelog noting what changed, and bump the version on line 155

borrrden (changes requested):
Also why did the default port situation get resolved?  I don't see a reply to what I said.  Currently the default is to let the system choose the port, and it will be set upon start.

servers/ios/TestServer/Utils/DatabaseManager.swift
borrrden: This will fail if a replicator is created twice with the same label, as is done in one of the tests.  I had requested something like this to indicate passing nil is the way to say "use the previous existing identity"

client/src/cbltest/api/listener.py
barkha06: I am not sure, but here's the failure: 
--> /Users/barkha.goyal/Desktop/temp2/couchbase-lite-tests/tests/QE/test_peer_to_peer.py:376:24 | 374 | all_dbs[1], 375 | endpoint=cblpytest.test_servers[0].replication_url( 376 | "db1", listener1.port, tls=True | ^^^^^^^^^^^^^^ Expected int, found Unknown | int | None 377 | ), 378 | replicator_type=replicator_type, | info: Element None of this union is not assignable to int info: Method defined here --> /Users/barkha.goyal/Desktop/temp2/couchbase-lite-tests/client/src/cbltest/api/testserver.py:146:9 | 144 | await self.__request_factory.send_request(self.__index, request) 145 | 146 | def replication_url(self, db_name: str, port: int, tls: bool = False): | ^^^^^^^^^^^^^^^ --------- Parameter declared here 147 | """

servers/ios/TestServer/Utils/DatabaseManager.swift
barkha06: Done

client/src/cbltest/api/listener.py
barkha06: One way to avoid would this is to make port a private parameter and define a getter for it, since the getter will always return an int, the ty lint would accept it.

barkha06:
> Also why did the default port situation get resolved? I don't see a reply to what I said. Currently the default is to let the system choose the port, and it will be set upon start.

I might've done that by mistake, I apologise for the oversight

spec/api/api.yaml
barkha06: Done

spec/tests/QE/test_peer_to_peer.md
```diff
@@ -0,0 +1,100 @@
+# Peer-to-Peer Tests
+
+This document describes the test cases for Couchbase Lite peer-to-peer URLEndpointListener replication, covering concurrent updates, various topology setups (one-to-many, many-to-one), multiple databases, and server failure/restart scenarios.
+
+## test_peer_to_peer_concurrent_replication
+
+Test peer-to-peer replication between a listener and a client, including concurrent document updates on both devices while replication is active.
+
+1. Verify CBL version >= 2.8.0 on all test servers.
```
borrrden: Not really a test step, as this won't fail the test it will only skip it.  Furthermore checking this is not really necessary because we don't have test server builds that old.

borrrden (approved):

spec/tests/QE/test_peer_to_peer.md
barkha06: Removed it

PR #386, opened 2026-04-15
CBG-5213 support dropping rosmar buckets

torcolvin:
- Create CouchbaseCloud.drop_bucket function since this might be a Sync Gateway operation or a server operation
- Change the behavior when creating the database to make sure all databases that are backed by a given bucket are actually gone. If you had multiple databases on the same bucket using different collections, then dropping all of them is necessary to avoid issues creating a database if the new databases targets different collections than the dropped one.
- Add special rosmar option to sync gateway config support dropping rosmar buckets
- Make sure to stop sync gateway if it was already running

Testing done:

```
uv run environment/local/start_local.py --build-testserver 4.0.3
uv run environment/local/run_sync_gateway.py --server rosmar --start
uv run pytest --config environment/local/rosmar_config.json tests/dev_e2e 
```

And with Couchbase Server:

```
uv run environment/local/run_sync_gateway.py --server cbs --start
uv run pytest --config environment/local/cbs_config.json tests/dev_e2e 
```

### DIFF
--- a/client/src/cbltest/api/cloud.py
+++ b/client/src/cbltest/api/cloud.py
@@ -159,10 +159,15 @@ class CouchbaseCloud:
 
             await self.__sync_gateway.load_dataset(dataset_name, data_filepath)
 
-    async def drop_bucket(self, bucket_name: str, *, wait_for_deleted=False):
-        """Drop the bucket from the backing cluster. This is an asynchronous operation unless wait_for_deleted is set to True."""
+    async def drop_bucket(self, bucket_name: str):
+        """Drop the bucket from the backing cluster."""
         if self.sync_gateway.using_rosmar:
-            await self.sync_gateway._send_request("delete", f"/_rosmar/{bucket_name}")
+            try:
+                await self.sync_gateway._send_request(
+                    "delete", f"/_rosmar/{bucket_name}"
+                )
+            except CblSyncGatewayBadResponseError as e:
+                if e.code != 404:
+                    raise
         else:
             self.couchbase_server.drop_bucket(bucket_name)
-            await self.couchbase_server.wait_for_bucket_deleted(bucket_name)
--- a/client/src/cbltest/api/cloud.py
+++ b/client/src/cbltest/api/cloud.py
@@ -129,14 +129,9 @@ class CouchbaseCloud:
 
                 current_span.add_event("Handle HTTP 412")
                 await self.__sync_gateway.delete_database(dataset_name)
-                if self.sync_gateway.using_rosmar:
-                    raise CblTestError(
-                        f"Database {dataset_name} already exists on Sync Gateway and cannot be deleted when "
-                        "using rosmar until CBG-5213 is implemented. To work around, restart a Sync Gateway to "
-                        "delete the rosmar buckets."
-                    )
-                else:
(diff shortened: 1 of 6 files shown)


client/src/cbltest/api/cloud.py
```diff
@@ -163,3 +158,11 @@ async def configure_dataset(
                 )
 
             await self.__sync_gateway.load_dataset(dataset_name, data_filepath)
+
+    async def drop_bucket(self, bucket_name: str, *, wait_for_deleted=False):
+        """Drop the bucket from the backing cluster. This is an asynchronous operation unless wait_for_deleted is set to True."""
```
borrrden: Is it?  I don't see anywhere that it is consulted....

borrrden (commented):
wait_for_deleted looks suspicious to me...

vipbhardwaj (approved):
LGTM 👍🏻

client/src/cbltest/api/cloud.py
torcolvin: This is part of a subsequent PR, and the code was buggy so I removed it.

client/src/cbltest/api/cloud.py
```diff
@@ -129,14 +129,9 @@ async def configure_dataset(
...
-                    raise CblTestError(
-                        f"Database {dataset_name} already exists on Sync Gateway and cannot be deleted when "
-                        "using rosmar until CBG-5213 is implemented. To work around, restart a Sync Gateway to "
-                        "delete the rosmar buckets."
-                    )
-                else:
-                    self.__couchbase_server.drop_bucket(db_payload.bucket)
+                await self.drop_bucket(db_payload.bucket)
+                await self.__sync_gateway.wait_for_no_databases(db_payload.bucket)
+                if not self.sync_gateway.using_rosmar:
```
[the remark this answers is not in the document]
torcolvin: When you drop a bucket, all the databases will be removed, so this is actually the correct behavior.

client/src/cbltest/api/syncgateway.py
```diff
@@ -1874,6 +1875,22 @@ async def start(self, config_name: str = "bootstrap") -> None:
...
+        retry=tenacity.retry_if_exception_type(AssertionError),
+    )
+    async def wait_for_no_databases(self, bucket_name: str):
+        with self._tracer.start_as_current_span("get_all_dbs"):
+            resp = await self._send_request("get", "/_all_dbs?verbose=true")
+            assert isinstance(resp, list), resp
+            for db in resp:
+                assert db["bucket"] != bucket_name, (
+                    f"Database {db=} is still backed by bucket {bucket_name}"
+                )
```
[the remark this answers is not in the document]
torcolvin: If the request ever fails for these reasons it _shouldn't_ be retried.

environment/local/run_sync_gateway.py
```diff
@@ -49,6 +49,7 @@ def main(start, stop, server):
     if stop:
         bridge.stop("localhost")
     else:
+        bridge.stop("localhost")
```
borrrden: This makes the if / else redundant here

borrrden (approved):
Probably just fix this tiny thing before merging

PR #387, opened 2026-04-15
CBG-5281: Fix use of SyncGatewayUserClient

torcolvin:
CBG-5281: Fix use of SyncGatewayUserClient

This takes care of code in https://github.com/couchbaselabs/couchbase-lite-tests/pull/383

Tested by running test_xattrs.py

### DIFF
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -529,7 +529,7 @@ class DatabaseStatusResponse:
 class _SyncGatewayBase:
     """
     Base class for Sync Gateway clients containing common document and database operations.
-    This class should not be instantiated directly - use SyncGateway or SyncGatewayPublic instead.
+    This class should not be instantiated directly - use SyncGateway or SyncGatewayUserClient instead.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


client/src/cbltest/api/syncgateway.py
```diff
@@ -1615,6 +1603,21 @@ def __init__(
...
+            timeout=10,
+        )
+        r.raise_for_status()
+        try:
+            self.using_rosmar = r.json()["bootstrap"]["server"].startswith("rosmar")
+        except AttributeError:
+            raise CblTestError(
+                "Unexpected response {r.json()} from Sync Gateway /_config endpoint, cannot determine if using Rosmar"
+            ) from None
+            self.using_rosmar = False
```
vipbhardwaj: This can actually be moved up, maybe even above the try/catch. Since by default we start-safe with a False for rosmar and then if the `r.json()` has a valid resp with all the nested fields there, it can assign  `True` to it. Since right now its unreachable anyways

client/src/cbltest/api/syncgateway.py
```diff
@@ -1615,6 +1603,21 @@ def __init__(
...
+            # disable hostname verification as we do in _create_session
+            verify=False,
+            timeout=10,
+        )
+        r.raise_for_status()
+        try:
+            self.using_rosmar = r.json()["bootstrap"]["server"].startswith("rosmar")
+        except AttributeError:
+            raise CblTestError(
+                "Unexpected response {r.json()} from Sync Gateway /_config endpoint, cannot determine if using Rosmar"
```
vipbhardwaj: Also this is intended to be a formatted string since the `{r.json()}` inside

client/src/cbltest/api/syncgateway.py
```diff
@@ -1615,6 +1603,21 @@ def __init__(
         """
         super().__init__(url, username, password, port, secure)
         self.__public_port = public_port
+        r = requests.get(
+            f"{self.scheme}{url}:{port}/_config",
+            auth=(username, password),
+            # disable hostname verification as we do in _create_session
+            verify=False,
+            timeout=10,
```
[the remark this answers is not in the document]
torcolvin: That's OK.

vipbhardwaj (approved):
LGTM, ran using this on the pipeline and the tests pass ✅

PR #388, opened 2026-04-16
fix: SGW 4.0.0 shown as 0.0.0 + keep all-fail runs

vipbhardwaj:
https://jira.issues.couchbase.com/browse/CBG-5293

SGW docs were added properly but the runs for 4.0.0 were shown as 0.0.0 because the build couldn't be fetched perhaps. Since the build might be fetched later, the current fix to support SGW and not trash other platforms is to keep version ONLY if `version_components` doesn't have two elements.
Which is the current SGW case

### DIFF
--- a/client/src/cbltest/greenboarduploader.py
+++ b/client/src/cbltest/greenboarduploader.py
@@ -63,7 +63,7 @@ class GreenboardUploader:
         :param platform: The platform name (e.g. couchbase-lite-net) as specified by the test server
         :param version: The version string (e.g. 3.2.0-b0136, etc) as specified by the test server
         """
-        if self.__overall_fail and not self.__has_sgw_marker:
+        if self.__overall_fail:
             cbl_warning("Overall result is failure, skipping upload...")
             return
 
--- a/client/src/cbltest/greenboarduploader.py
+++ b/client/src/cbltest/greenboarduploader.py
@@ -63,17 +63,26 @@ class GreenboardUploader:
         :param platform: The platform name (e.g. couchbase-lite-net) as specified by the test server
         :param version: The version string (e.g. 3.2.0-b0136, etc) as specified by the test server
         """
-        if self.__overall_fail:
+        if self.__overall_fail and not self.__has_sgw_marker:
             cbl_warning("Overall result is failure, skipping upload...")
             return
 
-        version_components = version.split("-")
-        if len(version_components) != 2:
-            version = "0.0.0"
-            build = 0
-        else:
-            version = version_components[0]
-            build = int(version_components[1].lstrip("b"))
+        version_to_parse = sgw_version if platform == "sync-gateway" else version
+        version_components = version_to_parse.split("-")
+
+        parsed_version = "0.0.0"
+        parsed_build = 0
+
+        if len(version_components) > 0 and version_components[0]:
+            parsed_version = version_components[0]
+
+        if len(version_components) > 1:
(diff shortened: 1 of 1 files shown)


torcolvin:
What's the root cause of the build showing 0.0.0 ? Can you write a test for the build types as copilot suggests?

One thing I noticed when doing some local testing is that `SyncGateway.get_version` will not be able to return a version number if it is a dev build https://github.com/couchbaselabs/couchbase-lite-tests/blob/28588c3a9daa5a945a2bbc20a8d7495b08ad2b63/client/src/cbltest/api/syncgateway.py#L647.

A Sync Gateway dev build should never get uploaded to greenboard so that isn't a problem itself, but I wonder where the root cause was.

client/src/cbltest/greenboarduploader.py
```diff
@@ -63,17 +63,26 @@ def upload(self, platform: str, os_name: str, version: str, sgw_version: str):
         :param platform: The platform name (e.g. couchbase-lite-net) as specified by the test server
         :param version: The version string (e.g. 3.2.0-b0136, etc) as specified by the test server
         """
-        if self.__overall_fail:
+        if self.__overall_fail and not self.__has_sgw_marker:
```
borrrden, line 66: Why is this being added?  It the tests failed to run properly the results should not be uploaded.

borrrden:
> A Sync Gateway dev build should never get uploaded to greenboard so that isn't a problem itself, but I wonder where the root cause was.

What does this mean?  The E2E tests are meant to validate an internal build.  Is that different than a dev build?  I didn't design this with sync gateway uploading in mind which means it will need a bit of thought to be able to retrieve what build number it is working with.

borrrden (changes requested):

vipbhardwaj:
This can be dropped since I have confirmed the fixes for rosmar in the PR you raised and the sgw-functional pipeline is passing ✅

client/src/cbltest/greenboarduploader.py
vipbhardwaj: Not really. 
If all tests fail, then should that be checked on Jenkins?
If yes, then what's the point of greenboard anyways? Use jenkins for all job results.

vipbhardwaj: CBL can keep it that way, hence I only added what would make SGW test show up on Jenkins if they still all failed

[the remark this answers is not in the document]
vipbhardwaj: `upload` is always called after the class is initialized, and hence the platform is always pre-determined.
So this is a good catch but didn't deep dive, if the class isn't initialized, platform isn't assigned, then `upload` isn't called and hence this `if` condition is never met.

TLDR;
The `__has_sgw_marker` will always have a determined value before we reach this LOC. But, good try "co-pilot"

borrrden: If I recall correctly this is only if the tests failed because of a problem setting them up or something along those those lines which is why I skipped the upload

borrrden: You can see above that overall fail is only set when report.when != call.  That means it failed for reasons other than failing tests, which means the run will just waste space in the results.

vipbhardwaj: Done ✅

borrrden (approved):

PR #390, opened 2026-04-20
fix: necessary additions for pipeline

vipbhardwaj:
[CBG-5300](https://jira.issues.couchbase.com/browse/CBG-5300)

Just updated the versions for CBL so that pipeline doesn't fail constantly.
Added a testserver since SGW tests do need now, it and was omitted before.

[CBG-5300]: https://couchbasecloud.atlassian.net/browse/CBG-5300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jenkins/pipelines/QE/sgw/topology.json
+++ b/jenkins/pipelines/QE/sgw/topology.json
@@ -1,6 +1,14 @@
 {
     "$schema": "../../../../environment/aws/topology_setup/topology_schema.json",
-    "test_servers": [],
+    "test_servers": [
+		{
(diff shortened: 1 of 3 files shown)


borrrden (approved):
This is going to get annoying REALLY fast.  How often are these called with the default values?  There is a good version reporter API to get the latest version of a product too so maybe that should be considered.

PR #394, opened 2026-04-22
Added 2 multiple client system test cases for Edge Server

sanjivanipatrax:
This PR adds in 2 multiple client system test cases for Edge Server which were remaining in Phase 3.

**Features added :-** 
- 2 system test cases, `test_system_multi_client_concurrent` and `test_system_multi_client_chaos`, in `tests/qe/edge_server/test_system.py`.
- Additional logging to the existing system test cases.

### DIFF
--- a/spec/tests/QE/edge_server/test_system.md
+++ b/spec/tests/QE/edge_server/test_system.md
@@ -6,13 +6,6 @@ This document describes the long-running system tests for Couchbase Lite Edge Se
 
 Test that create, update, and delete operations replicate correctly between Sync Gateway and Edge Server over an extended period (up to 6 hours), with cycle (sync_gateway or edge_server) and operations (create, create_update_delete, or create_delete) randomly chosen per document.
 
-**Parameters:**
-- Document counter starts at 11 and increments each iteration
-- `cycle`: Random choice from `["sync_gateway", "edge_server"]`
-- `operations`: Random choice from `["create", "create_update_delete", "create_delete"]`
-
-<br>
-
 **Steps:**
 1. **Setup**
    * Create a bucket `bucket-1` on Couchbase Server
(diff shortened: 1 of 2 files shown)


borrrden (changes requested):
Missing test specs

tests/QE/edge_server/test_system.py
```diff
@@ -128,6 +129,7 @@ async def test_system_one_client_l(
 
             if cycle == "sync_gateway":
                 # Create on Sync Gateway and validate on Edge Server
+                self.mark_test_step(f"Creating {doc_id} on Sync Gateway.")
```
borrrden, line 132: These test steps are supposed to mirror what the test steps are in the markdown.  Could you change one or the other so that they line up better.  Otherwise it's kind of hard to find which step in the markdown these correspond to.

borrrden (changes requested):

tests/QE/edge_server/test_system.py
sanjivanipatrax: @borrrden Would it be good if I add a step number for each so it's easier to correlate the test steps?

borrrden: The api itself automatically add step numbers so no.

spec/tests/QE/edge_server/test_system.md
```diff
@@ -1,70 +1,158 @@
 # System Tests (Edge Server)
 
-This document describes the long-running system tests for Couchbase Lite Edge Server replication and resilience, including a stability run and a chaos run with periodic Edge Server kill/restart.
+This document describes the long-running system tests for Couchbase Lite Edge Server replication and resilience. Tests cover both single-client and multi-client (4 concurrent async clients) scenarios, each in two modes: a stability run with steady CRUD traffic, and a chaos run that periodically kills and restarts the Edge Server.
+
+> Each *italicised* bullet phrase is the exact string passed to `mark_test_step` in the code, in the same order. Trailing text after the em-dash describes what happens at that step.
```
borrrden, line 5: This is a neat idea!

borrrden (approved):
If there is anything else you think will be useful with diagnosing, you can use the cbl_info and family of logging to put it into the logs, but as you've done here "mark test step" should be a touch point with the markdown so that someone unfamiliar with the test can see the reasoning of the test.

PR #396, opened 2026-04-23
Add databse

vipbhardwaj:
[CBG-5317](https://jira.issues.couchbase.com/browse/CBG-5317)

This PR is because I was not aware that long back there was a change where database_version was no longer a default 4.0 automatic value but had to be added as a parameter to be passed in : 
- Jenkinsfiles
- shell scripts

for every test script.
Hence this is the PR to update the same.

[CBG-5317]: https://couchbasecloud.atlassian.net/browse/CBG-5317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jenkins/pipelines/QE/upg-sgw/Jenkinsfile_rolling
+++ b/jenkins/pipelines/QE/upg-sgw/Jenkinsfile_rolling
@@ -2,7 +2,7 @@ pipeline {
     agent none
     parameters {
         string(name: 'CBL_VERSION', defaultValue: '3', description: 'Couchbase Lite Version (major only auto-resolves via ProGet)')
-        string(name: 'DATASET_VERSION', defaultValue: '4.0', description: 'Dataset Version to use for test-server')
+        string(name: 'DATASET_VERSION', defaultValue: '3.2', description: 'Dataset Version to use for test-server')
         string(name: 'SGW_VERSIONS', defaultValue: '3.2.7 3.3.3 3.3.4', description: 'Sync Gateway Versions to upgrade across')
     }
     stages {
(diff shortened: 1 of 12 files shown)


borrrden:
Why are you requesting review on a PR into another PR?  This is how things start getting crazy....

jenkins/pipelines/QE/upg-sgw/Jenkinsfile_rolling
```diff
@@ -19,6 +19,7 @@ pipeline {
     agent none
     parameters {
         string(name: 'CBL_VERSION', defaultValue: '3', description: 'Couchbase Lite Version (major only auto-resolves via ProGet)')
+        string(name: 'DATASET_VERSION', defaultValue: '4.0', description: 'Dataset Version to use for test-server')
```
borrrden: DATASET_VERSION default 4.0 for 3.x version is not a good default

vipbhardwaj:
How so?
Ultimately this will go to main so it should be reviewed. Hence I requested the review.
Does that have a caveat?

jenkins/pipelines/QE/upg-sgw/Jenkinsfile_rolling
vipbhardwaj: Fixed ✅

borrrden:
The caveat is if you merge it into the PR branch by mistake instead of main (or merge the other PR before this one), then it has to get re-reviewed for another PR.  If you want it to be in the PR just put it directly to the PR branch it is less confusing.

borrrden (approved):

vipbhardwaj:
I understand. So the right way is for waiting for the parent PR to get merged to main then requesting a review on this?

PR #397, opened 2026-04-23
Migrate remaining chaos/MTLS/CRUD tests for edge server

barkha06:
- Introduced CA-based TLS and mTLS support
- Added certificate generation utilities and updated ES setup to provision CA, server, and client certs
- Added mTLS test coverage
- Added chaos tests: test_edge_server_offline_sync_and_recovery and test_edge_server_with_concurrent_rest_requests
- Added Bulk doc CRUD test

### DIFF
--- a/environment/aws/common/x509_certificate.py
+++ b/environment/aws/common/x509_certificate.py
@@ -114,7 +114,7 @@ def create_signed_cert(cn: str, ca: CertKeyPair):
         .public_key(key.public_key())
         .serial_number(random_serial_number())
         .not_valid_before(datetime.now(timezone.utc))
-        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=1))
+        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=365))
         .add_extension(
             ExtendedKeyUsage([ExtendedKeyUsageOID.SERVER_AUTH]),
             critical=False,
@@ -137,7 +137,7 @@ def create_client_cert(cn: str, ca: CertKeyPair):
         .public_key(key.public_key())
         .serial_number(random_serial_number())
         .not_valid_before(datetime.now(timezone.utc))
-        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=1))
+        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=365))
         .add_extension(
             ExtendedKeyUsage([ExtendedKeyUsageOID.CLIENT_AUTH]),
             critical=False,
--- a/environment/aws/common/x509_certificate.py
+++ b/environment/aws/common/x509_certificate.py
@@ -9,6 +9,7 @@ from cryptography.hazmat.primitives.serialization import (
     pkcs12,
 )
 from cryptography.x509 import (
+    BasicConstraints,
     Certificate,
     CertificateBuilder,
     ExtendedKeyUsage,
@@ -79,3 +80,69 @@ def create_self_signed_certificate(CN: str) -> CertKeyPair:
     )
 
     return CertKeyPair(leaf_certificate, private_key)
+
+
+def create_ca():
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+
+    subject = issuer = Name([NameAttribute(NameOID.COMMON_NAME, "EdgeTestCA")])
+
+    cert = (
+        CertificateBuilder()
+        .subject_name(subject)
+        .issuer_name(issuer)
+        .public_key(key.public_key())
+        .serial_number(random_serial_number())
+        .not_valid_before(datetime.now(timezone.utc))
+        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=365))
+        .add_extension(BasicConstraints(ca=True, path_length=None), critical=True)
+        .sign(key, hashes.SHA256())
+    )
+
+    return CertKeyPair(cert, key)
+
+
+def create_signed_cert(cn: str, ca: CertKeyPair):
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+
+    subject = Name([NameAttribute(NameOID.COMMON_NAME, cn)])
+
+    cert = (
+        CertificateBuilder()
+        .subject_name(subject)
+        .issuer_name(ca.certificate.subject)
+        .public_key(key.public_key())
+        .serial_number(random_serial_number())
+        .not_valid_before(datetime.now(timezone.utc))
+        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=1))
+        .add_extension(
+            ExtendedKeyUsage([ExtendedKeyUsageOID.SERVER_AUTH]),
+            critical=False,
+        )
+        .sign(ca.private_key, hashes.SHA256())
+    )
+
+    return CertKeyPair(cert, key)
+
+
+def create_client_cert(cn: str, ca: CertKeyPair):
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+
+    subject = Name([NameAttribute(NameOID.COMMON_NAME, cn)])
+
+    cert = (
+        CertificateBuilder()
+        .subject_name(subject)
+        .issuer_name(ca.certificate.subject)
+        .public_key(key.public_key())
+        .serial_number(random_serial_number())
+        .not_valid_before(datetime.now(timezone.utc))
+        .not_valid_after(datetime.now(timezone.utc) + timedelta(days=1))
+        .add_extension(
+            ExtendedKeyUsage([ExtendedKeyUsageOID.CLIENT_AUTH]),
+            critical=False,
+        )
+        .sign(ca.private_key, hashes.SHA256())
+    )
+
+    return CertKeyPair(cert, key)
--- a/spec/tests/QE/edge_server/test_chaos_scenarios.md
+++ b/spec/tests/QE/edge_server/test_chaos_scenarios.md
@@ -33,4 +33,29 @@ Test multi-edge synchronization and recovery with chained replications.
 9. Kill Edge Server 1, update documents via Edge Servers 2/3, and verify replication.
 10. Kill Edge Servers 1 and 3, update documents via Edge Server 2, and verify replication.
 
+## test_edge_server_offline_sync_and_recovery
 
+Test the offline synchronization and recovery capabilities of a chained Edge Server topology.
+
+1. Configure Edge Server 1 to replicate from Sync Gateway for the `travel` dataset.
+2. Configure Edge Server 2 to replicate from Edge Server 1.
+3. Configure Edge Server 3 to replicate from Edge Server 2.
+4. Wait for replication to become idle across all Edge Servers.
+5. Delete all documents in the `travel.hotels` collection on Edge Server 3 using a bulk delete.
+6. Wait for replication to become idle, then kill Edge Server 3.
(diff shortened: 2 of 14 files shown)


spec/tests/QE/edge_server/test_chaos_scenarios.md
```diff
@@ -33,4 +33,29 @@ Test multi-edge synchronization and recovery with chained replications.
...
+2. Configure Edge Server 2 to replicate from Edge Server 1.
+3. Configure Edge Server 3 to replicate from Edge Server 2.
+4. Wait for replication to become idle across all Edge Servers.
+5. Delete all documents in the `travel.hotels` collection on Edge Server 3 using a bulk delete.
+6. Wait for replication to become idle, then kill Edge Server 3.
+7. Verify the documents are deleted in Edge Server 1, Edge Server 2, and Sync Gateway.
+8. Restart Edge Server 3 and wait for it to come online.
+9. Create 10,000 documents in `travel.hotels` via bulk create on Edge Server 3.
+10. Verify documents are created on Edge Server 3 and wait for replication to become idle.
+11. Kill Edge Server 3 again.
```
borrrden: What is the purpose of step 11?

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,253 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+            len(edge1_docs.rows) == len(sgw_docs.rows) == len(edge2_docs.rows) == 0
+        ), (
+            f"Collection hotels count mismatch in len(edge1_docs.rows): {len(edge1_docs.rows)} ,  len(sgw_docs.rows): {len(sgw_docs.rows)}, len(edge2_docs.rows):  {len(edge2_docs.rows)}"
+        )
+
+        self.mark_test_step("Start ES3")
+        await edge_server3.start_server()
+        await asyncio.sleep(100)
+
+        self.mark_test_step("Test document inserts")
```
borrrden: Not a great test step description, quite vague

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,253 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+        docgen = JSONGenerator(seed=10, size=10000)
+        create_docs = docgen.generate_all_documents()
+        bulk_ops = [
+            BulkDocOperation(body=doc, _id=id, optype="create")
+            for id, doc in create_docs.items()
+        ]
+        await edge_server3.bulk_doc_op(bulk_ops, "travel", "travel", "hotels")
+
+        self.mark_test_step("Verify document inserted")
```
borrrden: These test steps should try to mirror the markdown as much as possible so they are easier to find and reason about if you are looking for a certain step based on the markdown description.

borrrden (changes requested):
Overall the test steps should be a little more specific about what they are doing.  When I write the markdown, I try to have the steps in the markdown be identical to the test steps in the test.  This helps a lot with not having rot and accidentally drifting away from the spec intention.

spec/tests/QE/edge_server/test_chaos_scenarios.md
barkha06: It validates that the newly created documents have been fully propagated and stored across the replication.Also by killing ES3 again after replication reaches idle, we ensure no buffered data is lost. It's an attempt to test replication after source removal.

tests/QE/edge_server/test_chaos_scenarios.py
barkha06: yes makes sense, I'll modify them.

tests/QE/edge_server/test_chaos_scenarios.py
barkha06: Modified it.

tests/QE/edge_server/test_chaos_scenarios.py
barkha06: Done

spec/tests/QE/edge_server/test_chaos_scenarios.md
borrrden: It seems like the result would be the same without it, would it not?

tests/QE/edge_server/test_authentication.py
```diff
@@ -44,11 +44,19 @@ async def test_basic_auth(self, cblpytest: CBLPyTest) -> None:
...
+        self, cblpytest: CBLPyTest, dataset_path: Path
+    ) -> None:
         self.mark_test_step("test_valid_tls")
         edge_server = await cblpytest.edge_servers[0].configure_dataset(
             db_name="names", config_file=f"{SCRIPT_DIR}/config/test_tls_config.json"
         )
-        self.mark_test_step("get server information")
+        self.mark_test_step("get server information with TLS")
+        version = await edge_server.get_version()
+        self.mark_test_step(f"VERSION:{version}")
```
borrrden: Looks like it should be a normal log message, not a test step.

tests/QE/edge_server/test_authentication.py
```diff
@@ -44,11 +44,19 @@ async def test_basic_auth(self, cblpytest: CBLPyTest) -> None:
...
+        edge_server = await cblpytest.edge_servers[0].configure_dataset(
+            db_name="names", config_file=f"{SCRIPT_DIR}/config/test_mtls_config.json"
+        )
+        self.mark_test_step("get server information with TLS")
```
borrrden: With mTLS right?

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,267 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+                local_doc = docs_dict.get(doc_id)
+                if not local_doc:
+                    return True  # deleted concurrently
+
+                for key, value in remote_doc.body.items():
+                    assert local_doc.get(key) == value
+                return True
+
+        except Exception as e:
+            return {"error": str(e), "optype": optype, "doc_id": doc_id}
```
borrrden: The rest of this function returns boolean but this returns a dict.  That's strange, and how does it work with assert?

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,267 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+                op,
+                docs,
+                docgen,
+                db_name="db",
+                scope="",
+                collection="",
+                revmap=revmap,
+            )
+
+        self.mark_test_step("Final consistency validation")
```
borrrden: Should say what it is validating

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,267 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+        all_docs = await edge_server.get_all_documents(db_name="db")
+        assert len(all_docs.rows) == docgen.size
+
+        revmap = all_docs.revmap
+        operations = ["create", "update", "delete", "read"]
+
+        self.mark_test_step("Run randomized CRUD workload")
+        for i in range(1000):
+            op = random.choice(operations)
+            self.mark_test_step(f"Iteration {i}: {op}")
```
borrrden: Logging message, not test step

tests/QE/edge_server/test_chaos_scenarios.py
```diff
@@ -267,3 +268,267 @@ async def test_3_edge_with_sync(self, cblpytest, dataset_path) -> None:
...
+        docs = docgen.generate_all_documents()
+
+        # Seed database
+        bulk_ops = [
+            BulkDocOperation(body=doc, _id=doc_id, optype="create")
+            for doc_id, doc in docs.items()
+        ]
+        await edge_server.bulk_doc_op(bulk_ops, db_name="db")
+
+        self.mark_test_step("Verify initial document load")
```
borrrden: Verify how (i.e. write that it is checking the count against the expected count)

tests/QE/edge_server/test_crud.py
```diff
@@ -176,3 +177,66 @@ async def test_single_doc_crud_ttl(
             self.mark_test_step(
                 f"Deleted doc successfully threw exception on retrieval: {e}"
             )
+
+    @pytest.mark.asyncio(loop_scope="session")
+    async def test_multiple_doc_crud(
+        self, cblpytest: CBLPyTest, dataset_path: Path
+    ) -> None:
+        self.mark_test_step("test_multiple_doc_crud")
```
borrrden: Not a test step.  As a rule of thumb, each test step should have one or more corresponding asserts before the next one.

tests/QE/edge_server/test_crud.py
```diff
@@ -176,3 +177,66 @@ async def test_single_doc_crud_ttl(
...
+                BulkDocOperation(
+                    _id=all_docs.rows[i + 50].id,
+                    body={"rev": 2, "idx": i + 50},
+                    rev=all_docs.rows[i + 50].revid,
+                    optype="update",
+                )
+            )
+            updated.append(all_docs.rows[i + 50].id)
+
+        self.mark_test_step("Execute bulk document operations")
```
borrrden: This could combine all of the following test steps written here into one:  Execute bulk document operations, verify creations updates and deletions.

borrrden (changes requested):

spec/tests/QE/edge_server/test_chaos_scenarios.md
barkha06: Yes, do you think it would make more sense if i wait for ES2 (which replicates with ES3) to be idle and kill ES3 and then wait the other edge-servers to be idle

borrrden: It's not hurting anything I guess, but I don't see the point.  I am always wary of creating tests with too many steps.

barkha06: Okay then I'll remove the kill edge-server step

tests/QE/edge_server/test_chaos_scenarios.py
barkha06: when modifying the test from using the edge server class to create clients instead of HTTP clients class, I missed this step. Thanks for pointing it

environment/aws/common/x509_certificate.py
```diff
@@ -79,3 +80,69 @@ def create_self_signed_certificate(CN: str) -> CertKeyPair:
...
+            ExtendedKeyUsage([ExtendedKeyUsageOID.SERVER_AUTH]),
+            critical=False,
+        )
+        .sign(ca.private_key, hashes.SHA256())
+    )
+
+    return CertKeyPair(cert, key)
+
+
+def create_client_cert(cn: str, ca: CertKeyPair):
```
borrrden: These three methods and the existing one could be collapsed into a single method to avoid all of this code duplication.  I swear I've set this up at some point somewhere....maybe that was not in python though.  One method called `create_cert(cn: str, ca: CertKeyPair | None, is_ca: boolean, usages: list[ExtendedKeyUsageOID]` could handle all of these scenarios, with `create_self_signed_cert` just calling that, or removed and its invocation changed.

borrrden (changes requested):

PR #398, opened 2026-04-23
Make Sync Gateway version comparison robust for dev builds

torcolvin:
Goal: Allow code that uses `supports_version_vectors()` to pass with Sync Gateway dev builds, right now it throws an exception because `get_version` returns `"unknown"` for dev builds

Sync Gateway has some unfortunate behavior around reporting version numbers for dev builds. This is probably a bug in Sync Gateway but a dev build and a release build return different values.

For a dev build:

```
    "vendor": {
        "name": "Couchbase Sync Gateway",
        "version": "4.1"
    },
    "version": "Couchbase Sync Gateway/mybranch(9fb8653) EE"
```

For a release build:

```
    "vendor": {
        "name": "Couchbase Sync Gateway",
        "version": "4.0"
    },
    "version": "Couchbase Sync Gateway/4.0.4(8;release) EE"
```

Therefore, try to use `version` so that the version number has build number but fall back to `vendor.version` for dev builds. We probably never want to upload this to greenboard and if we did we'd want something to actually say the commit hash but that doesn't work with the build number being integer typed. I'm not trying to fix that problem in this ticket though.

@borrrden, looking for your input on whether you are OK using pydantic for json parsing.
@vipbhardwaj, I ran most of the QE tests but you should make sure this code works before I break Jenkins by mistake.

### DIFF
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -678,7 +678,7 @@ class _SyncGatewayBase:
             return ret_val
 
     async def supports_version_vectors(self) -> bool:
-        """Returns whether the Sync Gateway instance supports version vectors (i.e. is 3.0 or later)"""
+        """Returns whether the Sync Gateway instance supports version vectors (i.e. is 4.0 or later)"""
         version = await self.get_version()
         return packaging.version.parse(version.version) >= packaging.version.parse(
             "4.0"
@@ -692,7 +692,8 @@ class _SyncGatewayBase:
         # In a dev build /_status "version" does not contain major.minor so use model.vendor.version
         if model.vendor.version in model.version:
             sg_version = SyncGatewayVersion(model.version)
-        sg_version = SyncGatewayVersion(model.vendor.version)
+        else:
+            sg_version = SyncGatewayVersion(model.vendor.version)
         assert packaging.version.parse(sg_version.version), (
             "Failed to parse Sync Gateway version from /_status response {model}"
         )
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -694,9 +694,13 @@ class _SyncGatewayBase:
             sg_version = SyncGatewayVersion(model.version)
         else:
             sg_version = SyncGatewayVersion(model.vendor.version)
-        assert packaging.version.parse(sg_version.version), (
-            "Failed to parse Sync Gateway version from /_status response {model}"
-        )
+        try:
+            packaging.version.parse(sg_version.version)
+        except packaging.version.InvalidVersion as exc:
+            raise CblTestError(
+                "Failed to parse Sync Gateway version from /_status response: {resp}\n"
+                f"version={model.version}"
+            ) from exc
         return sg_version
 
     def tls_cert(self) -> str | None:
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -7,11 +7,13 @@ from pathlib import Path
 from typing import Any, cast
 from urllib.parse import urljoin
 
+import packaging.version
 import requests
 import tenacity
 from aiohttp import BasicAuth, ClientError, ClientSession, ClientTimeout, TCPConnector
 from aiohttp.client_exceptions import ClientConnectorError
 from opentelemetry.trace import get_tracer
+from pydantic import BaseModel
 
 from cbltest.api.error import CblSyncGatewayBadResponseError, CblTestError
 from cbltest.api.jsonserializable import JSONDictionary, JSONSerializable
@@ -486,6 +488,9 @@ class CouchbaseVersion(ABC):
         self.__version = parsed[0]
         self.__build_number = parsed[1]
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self.__raw})"
+
 
 class SyncGatewayVersion(CouchbaseVersion):
     """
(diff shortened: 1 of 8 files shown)


client/src/cbltest/api/syncgateway.py
```diff
@@ -635,17 +664,26 @@ async def _send_request(
...
+
+    async def get_version(self) -> SyncGatewayVersion:
+        """Return version of Sync Gateway"""
+        resp = await self._send_request("get", "/_status")
+        assert isinstance(resp, dict)
+        model = SyncGatewayStatusResponse.model_validate(resp)
+        # In a dev build /_status "version" does not contain major.minor so use model.vendor.version
+        if model.vendor.version in model.version:
+            sg_version = SyncGatewayVersion(model.version)
+        sg_version = SyncGatewayVersion(model.vendor.version)
```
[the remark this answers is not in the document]
torcolvin: This is done in SynGatewayVersion.parse

vipbhardwaj (approved):
LGTM

client/src/cbltest/api/syncgateway.py
```diff
@@ -486,32 +488,56 @@ def __init__(self, input: str):
...
 
     def parse(self, input: str) -> tuple[str, int]:
-        first_lparen = input.find("(")
-        first_semicol = input.find(";")
-        if first_lparen == -1 or first_semicol == -1:
-            return ("unknown", 0)
-
-        version = input[0:first_lparen].strip()
-        if not version:
+        m = re.match(r"^[^(\n]+", input)
```
borrrden: To spare us parsing regex manually, perhaps a comment could be here saying what we are looking for

client/src/cbltest/api/syncgateway.py
```diff
@@ -486,32 +488,56 @@ def __init__(self, input: str):
...
+        if m:
+            version = m.group().strip()
+        else:
             cbl_warning(f"Could not extract version from SGW version string: '{input}'")
             version = "unknown"
 
-        try:
-            build = int(input[first_lparen + 1 : first_semicol])
-        except ValueError:
+        m = re.search(r"(?<=\()([^;)]+)", input)
```
borrrden, line 518: Same for here

client/src/cbltest/api/syncgateway.py
```diff
@@ -648,23 +680,31 @@ async def _send_request(
...
+        return packaging.version.parse(version.version) >= packaging.version.parse(
+            "4.0"
+        )
+
+    async def get_version(self) -> SyncGatewayVersion:
+        """Return version of Sync Gateway"""
+        resp = await self._send_request("get", "/_status")
+        assert isinstance(resp, dict)
+        model = SyncGatewayStatusResponse.model_validate(resp)
+        # In a dev build /_status "version" does not contain major.minor so use model.vendor.version
```
borrrden: I'd like to flesh out this comment a bit more


```suggestion
        # Using model.version is preferable, but only if it contains at least the major.minor that is in the model.vendor.version key (dev builds don't have this)
```

client/src/cbltest/api/syncgateway.py
```diff
@@ -581,6 +608,11 @@ def port(self) -> int:
         """Gets the HTTP API port of the Sync Gateway instance"""
         return self.__port
 
+    @property
+    def public_port(self) -> int:
+        """Gets the public API port of the Sync Gateway instance"""
+        return self.__public_port
```
borrrden: If there is no need for using this property in tests then I'd like to stick with the "less is better" approach to the API.  So far it seems like the only piece of information related to this is the replication URL.

borrrden (changes requested):

client/src/cbltest/api/syncgateway.py
torcolvin: Wrote maybe too many comments because it turns out this was slightly wrong! I added some more tests.

client/src/cbltest/api/syncgateway.py
torcolvin: This is used in a way I think is perhaps silly because of https://github.com/couchbaselabs/couchbase-lite-tests/blob/df44841cfc7439d04d7c944b492d7396772b2c65/client/src/cbltest/api/syncgateway.py#L1358 but this is out of scope for this PR. It probably would be to remove this function and use `SyncGatewayUserClient`

torcolvin:
I made an accidental mistake by testing not quite prod versions of Sync Gateway from latestbuilds, so I would still love if someone can double check that a test like test_offline_processing_of_external_updates is not broken.

borrrden:
That tests still works but I had to disable TLS verification.  It seems like it has recently gotten stricter.  The CA cert is going to have to be regenerated at some point to include proper extensions.

borrrden (approved):

torcolvin:
> That tests still works but I had to disable TLS verification. It seems like it has recently gotten stricter. The CA cert is going to have to be regenerated at some point to include proper extensions.

When you say disable TLS verification, where did you do that? I don't make any new calls, I just use `resp = await self._send_request("get", "/_status")` 

Do you mean disabling TLS here https://github.com/couchbaselabs/couchbase-lite-tests/blob/62153cbeb61467e85d565bb78ba1697d9bfb5829/client/src/cbltest/api/syncgateway.py#L594 ? If you wanted to do that this seems like a separate PR.

borrrden:
> > That tests still works but I had to disable TLS verification. It seems like it has recently gotten stricter. The CA cert is going to have to be regenerated at some point to include proper extensions.
> 
> When you say disable TLS verification, where did you do that? I don't make any new calls, I just use `resp = await self._send_request("get", "/_status")`
> 
> Do you mean disabling TLS here
> 
> https://github.com/couchbaselabs/couchbase-lite-tests/blob/62153cbeb61467e85d565bb78ba1697d9bfb5829/client/src/cbltest/api/syncgateway.py#L594
> ? If you wanted to do that this seems like a separate PR.

I meant the latter.  The CA cert in the repo doesn't have extensions properly set and so is getting rejected by python 3.13.  It definitely is a separate PR to handle that (it means recreating the CA cert and leaf certs)

PR #400, opened 2026-04-27
Parametrize Multipeer QE tests to run mixed-mode testing

barkha06:
* Added a build_transport_group function to create a mix of wifi-only, bluetooth- only, and All as transport types for replication to test mixed-mode
* Changed getMultipeerReplicatorStatus response type to be parsed using _assert_string_entry since Android does not return a transport type yet.
* Parametrized document counts and timeouts based on transport type in multipeer tests

### DIFF
--- a/client/src/cbltest/api/multipeer_replicator_types.py
+++ b/client/src/cbltest/api/multipeer_replicator_types.py
@@ -33,40 +33,6 @@ class MultipeerTransportType(Flag):
             if m.value != 0 and (m.value & (m.value - 1)) == 0 and (self & m) == m
         ]
 
-    @classmethod
-    def build_group_transports(
-        cls,
-        size: int,
-        transport: str,
-    ) -> list["MultipeerTransportType"]:
-        """
-        Builds a list of transports for a group of given size.
-
-        If transport != "MIXED_MODE", returns a uniform list.
-        If MIXED_MODE, distributes WIFI, BLUETOOTH, and ALL.
-        """
-
-        # Non-mixed mode → uniform assignment
-        if transport != "MIXED_MODE":
-            return [cls.from_string(transport)] * size
-
-        if size == 1:
-            return [cls.ALL]
-
-        if size == 2:
-            return [cls.BLUETOOTH, cls.ALL]
-
-        # General distribution
-        num_wifi = size // 3
-        num_bt = size // 3
-        num_dual = size - num_wifi - num_bt
-
-        transports_array = (
-            [cls.WIFI] * num_wifi + [cls.BLUETOOTH] * num_bt + [cls.ALL] * num_dual
-        )
-
-        return transports_array
-
 
 class MultipeerReplicatorAuthenticator(JSONSerializable):
     """
--- a/client/src/cbltest/api/multipeer_replicator_types.py
+++ b/client/src/cbltest/api/multipeer_replicator_types.py
@@ -33,6 +33,40 @@ class MultipeerTransportType(Flag):
             if m.value != 0 and (m.value & (m.value - 1)) == 0 and (self & m) == m
         ]
 
+    @classmethod
+    def build_group_transports(
+        cls,
+        size: int,
+        transport: str,
+    ) -> list["MultipeerTransportType"]:
+        """
+        Builds a list of transports for a group of given size.
+
+        If transport != "MIXED_MODE", returns a uniform list.
+        If MIXED_MODE, distributes WIFI, BLUETOOTH, and ALL.
+        """
+
+        # Non-mixed mode → uniform assignment
+        if transport != "MIXED_MODE":
+            return [cls.from_string(transport)] * size
+
+        if size == 1:
+            return [cls.ALL]
+
+        if size == 2:
+            return [cls.BLUETOOTH, cls.ALL]
+
+        # General distribution
+        num_wifi = size // 3
+        num_bt = size // 3
+        num_dual = size - num_wifi - num_bt
+
+        transports_array = (
+            [cls.WIFI] * num_wifi + [cls.BLUETOOTH] * num_bt + [cls.ALL] * num_dual
+        )
+
+        return transports_array
+
 
 class MultipeerReplicatorAuthenticator(JSONSerializable):
     """
--- a/tests/QE/test_multipeer.py
+++ b/tests/QE/test_multipeer.py
@@ -93,6 +93,8 @@ class TestMultipeer(CBLTestClass):
         )
         expected_keys = {f"device{i + 1}" for i in range(len(all_dbs))}
         retry = 5
+        for i, doc in enumerate(results, 1):
+            print(f"device_{i}: {doc}")
         for doc in results:
             counter = doc.body["counter"]
             assert set(counter.keys()) == expected_keys, (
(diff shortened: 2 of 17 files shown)


client/src/cbltest/api/multipeer_replicator_types.py
```diff
@@ -33,6 +33,40 @@ def to_json(self) -> list[str]:
             if m.value != 0 and (m.value & (m.value - 1)) == 0 and (self & m) == m
         ]
 
+    @classmethod
+    def build_group_transports(
+        cls,
+        size: int,
+        transport: str,
```
borrrden: This creates a makeshift enum out of a magic string for no reason.  MultipeerTransportType is already a flag enum, which means it can be bitwise or together (MultipeerTransportType.BLUETOOTH | MultipeerTransportType.WIFI).  

Also this seems like a helper method specific to a test suite and not an API method.

tests/QE/test_system_multipeer.py
```diff
@@ -197,22 +208,19 @@ async def stop_and_restart(idx):
...
-    # simultaneously inserting 10000 documents and blobs different docs to each CBL and starting replication
     @pytest.mark.asyncio(loop_scope="session")
     @pytest.mark.parametrize(
-        "transport",
-        [
-            MultipeerTransportType.BLUETOOTH,
-            MultipeerTransportType.WIFI,
-        ],
+        "transport, doc_count",
+        [("BLUETOOTH", 10000), ("WIFI", 100000), ("MIXED_MODE", 50000)],
```
borrrden, line 212: What is the purpose of testing "mixed mode" in cases like this?  It's just going to be the same as wifi in the absence of any network factors.  It's not like it uses both transports at the same time, it's going to pick the best one and use that.

borrrden (changes requested):

tests/QE/test_system_multipeer.py
barkha06: It's to test peer-to-peer with a mix of devices using only wifi, devices using only bluetooth and a mix of both such that docs are being replicated between wifi-only device<> wifi+bt changes <> BT changes. There is a bug discovered with mixed mode testing as well([CBL-8138](https://jira.issues.couchbase.com/browse/CBL-8138))

client/src/cbltest/api/multipeer_replicator_types.py
barkha06: It's a test helper to help generate a mixed mode set with even distribution of wifi only, BT only and wifi+bluetooth devices. I need it for both test_multipeer.py and test_system_multipeer.py so I assumed it could go in the API to make it easy to import.
I've used strings for ease-of-use since the test parameter could of of type MultipeerTransportType or string to represent mixed-mode making it two different types of parameters. 
Do you think there's a better way to do this?

borrrden: These are flag style enums though so they can represent both anyway like I mentioned above.  It doesn't seem like it is the class's job to create collections of itself though, that seems like a higher level thing.

tests/QE/test_system_multipeer.py
borrrden: Right I forgot mixed mode means "different between device" not "using both modes"

client/src/cbltest/api/multipeer_replicator_types.py
barkha06: Where do you recommend I place this function?

borrrden: If it were me, I'd start a collection of test helper functions inside of the tests/QE folder.

borrrden: As an update, Vipul has now started a shared logic section that you can add this to.

barkha06: Yeah sure, I'll move it there, thanks

tests/QE/test_multipeer.py
```diff
@@ -9,26 +9,24 @@
...
 from cbltest.api.replicator_types import (
     ReplicatorCollectionEntry,
     ReplicatorConflictResolver,
 )
 from cbltest.api.test_functions import compare_doc_results_p2p
+from shared.multipeer_test_helpers import build_group_transports
 
 
 @pytest.mark.cbl
 @pytest.mark.min_test_servers(2)
```
borrrden, line 21: I think the minimum must be more than 2 in mixed mode.  What happens if the modes are bluetooth and wifi?  Nothing will ever sync....

tests/QE/test_multipeer.py
```diff
@@ -103,36 +101,44 @@ async def test_scalable_conflict_resolution(self, cblpytest: CBLPyTest, transpor
...
+
+        retry = 5
+        while retry > 0:
+            rev_ids = [tuple(sorted(doc.revs.split(","))) for doc in results]
+            if len(set(rev_ids)) == 1:
+                break
+            self.mark_test_step(
+                f"Rev IDs don't match across devices, waiting 30 seconds (retries left: {retry})"
+            )
+            await asyncio.sleep(30)
```
borrrden: 30 seconds again?!  This is part of the reason the previous framework ended up with DAYS of runtime fyi.  Casually throwing in a 30 second delay with no justification is not ok.

tests/QE/test_multipeer.py
```diff
@@ -755,50 +768,66 @@ async def test_large_document_replication(self, cblpytest: CBLPyTest, transport)
...
+                    for device_idx, docs in enumerate(all_docs_results[1:], 2):
+                        assert compare_doc_results_p2p(
+                            all_docs_results[0]["_default._default"],
+                            docs["_default._default"],
+                        ), f"Device {device_idx} content does not match device 1"
+                    success = True
+                except Exception as e:
+                    self.mark_test_step(f"Document validation failed: {e}")
+                    retry -= 1
+                    await asyncio.sleep(30)
```
borrrden: Another 30 second delay....

borrrden: Furthermore, this is pretty much guaranteed to run the full 5 times (= 150 seconds) if the previous wait for idle clause failed.

borrrden (changes requested):

tests/QE/test_multipeer.py
barkha06: The build_group_transports sets one of the transport type to ALL and the other to Bluetooth for size of array set to 2

tests/QE/test_multipeer.py
barkha06: The test step mentions that after resolving conflicts the revision IDs don't match so there is a retry of 30 seconds. Earlier the tests failed at this stage, since the mesh size is 15, 30 seconds is a very reasonable amount of time. This fix has been recommended by Pasin.

tests/QE/test_multipeer.py
barkha06: Yeah this can be removed since the test parameters have a sufficient timeout.

tests/QE/test_multipeer.py
borrrden: I talked to Pasin and he said 10 seconds would be ok as a retry interval.

PR #402, opened 2026-04-29
feat: SGW upg distinction for greenboard + date in schema for all tests

vipbhardwaj:
[CBG-5348](https://jira.issues.couchbase.com/browse/CBG-5348)

The above ticket has a very comprehensive explanation of WHAT and WHY is done in the PR.

[CBG-5348]: https://couchbasecloud.atlassian.net/browse/CBG-5348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenkins/pipelines/QE/upg-sgw/upload_upgrade_results.py
```diff
@@ -0,0 +1,130 @@
...
+This script is called once at the end of an upgrade test pipeline
+(test.sh / test_rolling.sh) after all pytest invocations have completed.
+
+It reads the deferred results file (one JSON line per pytest session),
+determines overall pass/fail, and uploads a single greenboard document
+for the entire upgrade batch.
+
+Environment variables (required):
+    SGW_UPGRADE_RESULTS_FILE  — path to the JSONL results file
+    SGW_UPGRADE_VERSIONS      — comma-separated ordered version list (e.g. "3.1.0,3.2.0,3.3.0")
```
borrrden: Why not just use regular arguments from the command line?  Then their usafe and notes will show up in the generated `--help` command.

client/src/cbltest/greenboarduploader.py
```diff
@@ -8,12 +11,20 @@
...
-    A class for uploading results to a specified greenboard server bucket
+    A class for uploading results to a specified greenboard server bucket.
+
+    Supports two modes:
+    - **Normal mode**: uploads results directly at end of pytest session.
+    - **Deferred mode** (upgrade jobs): writes pass/fail to a shared JSONL file
+      so a final script can upload one combined document per upgrade batch.
+
+    Deferred mode activates when ``SGW_UPGRADE_RESULTS_FILE`` is set in the
+    environment.
```
borrrden: I'd prefer this not be handled by environment variables.  You can add command line options at the end of the fixture python source.

client/src/cbltest/greenboarduploader.py
```diff
@@ -126,6 +160,51 @@ def upload(
...
+        target_version = upgrade_path[-1] if upgrade_path else "0.0.0"
+
+        self._upload_document(
+            {
+                "build": target_build,
+                "version": target_version,
+                "sgwVersion": target_sgw_version,
+                "upgradePath": upgrade_path,
+                "failCount": 0 if passed else 1,
+                "passCount": 1 if passed else 0,
```
borrrden: Is this the best way to show this information?  How will you know which version failed to upgrade?

borrrden (changes requested):

client/src/cbltest/greenboarduploader.py
vipbhardwaj: Does greenboard upgrade jobs need showing that as well?
> PS : On this note I was also thinking of somehow adding hyperlinks to each bar for the run it relates to, that way each (failing) bar is clickable to the jenkins run and hence a jump to its logs/params/etc

jenkins/pipelines/QE/upg-sgw/upload_upgrade_results.py
vipbhardwaj: I did this since upgrade jobs are upgrade jobs by default, they don't need us passing/telling them by CLI that "you're an upgrade job so behave that way when you reach greenboard fixture stage post your test run".
Right? I mean, They ARE upgrade jobs, so they don't need telling. A CLI arg with default value seems redundant.
Plus its kind of a hassle to then skip them and reach the SGW versions specified post them, in the end, to be iterated over.
But if you think that's the best/graceful way, I don't mind implementing it.

client/src/cbltest/greenboarduploader.py
borrrden: A bar either entirely red or entirely green seems not too great of a UI experience, is what I am getting at.  If they were separate then you'd be able to quickly examine that the upgrade from X to Y failed.  At this rate all you have is a binary red or green blob.

jenkins/pipelines/QE/upg-sgw/upload_upgrade_results.py
borrrden: It looks like you've just done the same thing as you described anyway, due to the fact that you have required environment variables.  I don't really understand the second half of this paragraph, all I am suggesting is that CLI arguments to this script are more graceful than environment variables.  The script itself would behave exactly the same.

client/src/cbltest/greenboarduploader.py
vipbhardwaj: Oh so separate bars for each upgrade iteration. That is a good idea indeed

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -9,6 +9,9 @@
 # fixture that will upload test results to greenboard, if it is
 # properly set up in config.json (see the schema for that file)
 # and if the --no-result-upload flag is not set on the command line.
+#
+# For upgrade jobs (SGW_UPGRADE_VERSIONS is set), each pytest session
```
borrrden, line 13: This could use some correction since it's not SGW_UPGRADE_VERSIONS anymore (same in other places)

client/src/cbltest/greenboarduploader.py
```diff
@@ -126,6 +122,75 @@ def upload(
...
+        target_version = upgrade_path[-1] if upgrade_path else "0.0.0"
+        target_build = 0
+        current_version = target_version
+        if sgw_version is not None:
+            current_version = sgw_version.version or target_version
+            target_version = upgrade_path[-1] if upgrade_path else current_version
+            target_build = sgw_version.build_number
+
+        # Infer upgradeFrom by finding current version in the path
+        upgrade_from = "initial"
```
borrrden, line 158: Is there a situation in which you won't find this?  Because upgradeFrom == initial doesn't seem meaningful.

borrrden (approved):

PR #405, opened 2026-05-05
Disable verify_lfs_checkout on linux

pasin:
Linux has the same problem as windows. The checkout field is false why the lfs files (dataset files) are checked out correctly. I need this now to be able to verify the RC libraries.

### DIFF
--- a/client/src/cbltest/utils.py
+++ b/client/src/cbltest/utils.py
@@ -1,6 +1,7 @@
 import json
 import os
 import subprocess
+import sys
 import time
 from collections.abc import Callable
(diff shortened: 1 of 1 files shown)


torcolvin:
Do you have logs of the failure? I find this check helpful for my team who doesn't use git lfs regularly but obviously this isn't worthwhile. That said, if I can have my cake and eat it too, I could figure out why this check doesn't work!

pasin:
@torcolvin Here is the log : http://jenkins.mobiledev.couchbase.com/job/staging-e2e-test-c/291/stages/log?nodeId=126

PR #414, opened 2026-05-15
refactor: redesign SGW upgrades + greenboard uploads

vipbhardwaj:
[CBG-5387](https://jira.issues.couchbase.com/browse/CBG-5387)

Redesigning of the Upgrade tests for SGW. Now upgrade tests work like a one shot source to target upgrade rather than a linear array of versions to keep upgrading through till reaching target. 

The end goal in a different PR would be to wrap this inside a list of versions where all permutations would run for each version as source and would upgrade to each major version.
That can be changed as discussed and agreed upon. 

**Please refer to the above ticket to see how the upgrades are technically designed to upgrade now.**

[CBG-5387]: https://couchbasecloud.atlassian.net/browse/CBG-5387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/src/cbltest/greenboarduploader.py
```diff
@@ -133,179 +134,78 @@ def upload(
...
-        target_version = (
-            upgrade_path[-1]
-            if upgrade_path
-            else iterations[-1].get("upgradeTo", "0.0.0")
-        )
+        versions = list({*doc.get("versions", []), from_version, to_version})
+        try:
+            versions.sort(key=Version)
+        except InvalidVersion:
+            cbl_warning(f"Falling back to lexicographic sort for versions {versions}")
```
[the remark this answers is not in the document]
vipbhardwaj: ✅

client/src/cbltest/greenboarduploader.py
```diff
@@ -133,179 +134,78 @@ def upload(
...
-            cbl_warning(f"Could not parse upgrade results file {path}: {e}")
-            return
-
-        iterations = state.get("iterations", [])
-        if not iterations:
-            cbl_warning(
-                f"Upgrade results file {path} has no iterations; skipping upload"
-            )
-            return
+            doc = coll.get(doc_id).content_as[dict]
```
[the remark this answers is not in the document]
vipbhardwaj: Mostly two different jobs won't run in parallel since the agent used only has one executor

tests/QE/test_rolling_upgrade_sgw.py
```diff
@@ -60,17 +59,16 @@ async def test_rolling_upgrade_sgw_cluster(
...
         num_docs_per_iteration = 10
         sg_db = "rolling_upg_db"
         bucket: str = SGW_BUCKET
         cbl_db = "rolling_upg_cbl"
         scope = "_default"
         collection = "_default"
 
-        self.mark_test_step(f"Phase: {upgrade_phase} | SGW version: {current_version}")
-        doc_id_prefix = f"rolling_{upgrade_phase}_{current_version}"
+        self.mark_test_step(f"SGW version: {current_version}")
```
borrrden: This is not a test step, this is a logging message

jenkins/pipelines/QE/upg-sgw/test.sh
```diff
@@ -72,67 +30,44 @@ TOPOLOGY_FILE="$AWS_ENVIRONMENT_DIR/topology_setup/topology.json"
...
-    export SGW_VERSION_UNDER_TEST="$CURRENT_SGW_VERSION"
-    export SGW_UPGRADE_PHASE="upgrade_to_$CURRENT_SGW_VERSION"
-    export SGW_PREVIOUS_VERSION="$PREVIOUS_SGW_VERSION"
-    pushd $QE_TESTS_DIR > /dev/null
-    uv run pytest -s -v --no-header -W ignore::DeprecationWarning --config config.json -m upg_sgw \
-        --upgrade-versions "$UPGRADE_VERSIONS" \
-        test_upg_sgw.py
-    popd > /dev/null
-done
+echo ">>> Post-upgrade pytest (uploads pass/fail to sgw-upgrade::waterfall doc) ..."
```
borrrden: I really think stuffing everything into a single document is an anti pattern and goes against the NoSQL way of doing things.  There is an excessive amount of special casing for this test when in reality it doesn't look that different from the other tests now.

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -8,12 +8,17 @@
...
-# fixture that will upload test results to greenboard, if it is
-# properly set up in config.json (see the schema for that file)
-# and if the --no-result-upload flag is not set on the command line.
+# fixture that will upload test results to greenboard, if it is properly
+# set up in config.json (see the schema for that file) and if the
+# --no-result-upload flag is not set on the command line.
 #
-# For upgrade jobs (SGW_UPGRADE_VERSIONS is set), each pytest session
-# uploads its own per-step result directly under platform="sgw-upgrade".
+# Two upload paths:
```
borrrden: This is the limit to which I will accept this branching.  If you need to add more modes then I am going to insist on a refactor because this is getting unwieldy.  Uploading to greenboard should be a simple thing, and the fact that it is not is a smell.

borrrden (changes requested):
Double check the test steps to make sure they are not simply logging messages in disguise.  They are meant to closely, if not identically, follow the markdown spec testing steps.  Furthermore I think stuffing everything into a single document is not a good way to go, but I won't push back against it too much now.  NoSQL is good for creating a bunch of small entries and aggregating them and the fact that they all need to be stuffed into a single document means that there is a lot of complicated branching happening purely in the greenboard uploader.  The more special cases there are, the more the code smells.  There needs to be some difference because of the information presented, but this is creating dozens of lines dedicated to just uploading sync gateway upload results.

jenkins/pipelines/QE/upg-sgw/test.sh
vipbhardwaj: Would you rather me create different documents just for the results of different upgrade tests results? Each time? That would actually create a lot of documents with just very very small information like the `source_version`, `target_version` and `result_of_the_test`. Mostly, plus on the frontend there would be extra time taken on aggregation of all the documents which belong to the SGW upgrade category and then show a result of all of those in a matrix of sorts.

Its more efficient to have a single document do the same, would almost never get over-populated since,
- For every repeated source and target run, it would just overwrite the previous pass/fail result.
- For every new source-target upgrade run, just a single extra object is added with 3 fields : `source`, `target`, `result`.
- For UI its much more compact, quick and easier to aggregate the results to be shown for "Upgrade tests SGW" when it just has to read a single document.

This was my motivation to stray away from the "New document for every test" approach, since a document for a stale `source,target` would not even matter and would lead to garbage/obsolete documents in the **greenboad** bucket. This last point was discussed with @bbrks as well.

tests/QE/test_rolling_upgrade_sgw.py
vipbhardwaj: Done

client/src/cbltest/plugins/greenboard_fixture.py
vipbhardwaj: I can refactor now, in this PR. Since no promises for future. Point me in a direction as to how?

jenkins/pipelines/QE/upg-sgw/test.sh
bbrks: My understanding was we'd have a single document **_for a given combination_** (e.g. `source,target`) which we can have a short array of results for (e.g. the last 5 runs for a given source->target upgrade)

This lets us constrain run history in an easy way, by limiting that array length and dropping the oldest element, but still lets us keep around as many `source,target` combinations as required.

That way, we can prune away EOL versions easily by just deleting by version e.g "2.8.3->3.0.4"

I don't know this code well at all to know what's going on, but it sounds like we now just have a single upgrade test result document that stores everything? Is that true?

client/src/cbltest/plugins/greenboard_fixture.py
borrrden: Split the code into two or three classes (one for default, and one or two for sgw waterfall/rolling).  Then have a factory choose between them and delegate the logic there.

jenkins/pipelines/QE/upg-sgw/test.sh
torcolvin: I do not think that we need _any_ branching code for the OS upgrade. We care about:

- What version of Sync Gateway are we trying to test? i.e. 4.1.0-1 build

Then when we run the tests, we only care about a set of tests, that are including but not limited to these types:

1. test_foo
2. test_bar
3. test_3.1.0->current
4. test_3.2.0->current
5. test_3.3.0->current
6. test_3.3.0->test_4.0.0->current
7. ....

These themselves are just entries in a pass/fail count under a single version. From my perspective, these can be on either represented that looks like "SGW 4.1.0-8" (this name is inexact) which could be the same column as all the normal non upgrade tests, or if this isn't possible since the pass/fail counts originate from a single Jenkins run.

I'd imagine that there are 2 different Jenkins pipeline jobs for this, and each will update the same platof with pass/fail counts. If there's a machine failure, we do want to be able to wipe out a set of runs and rerun it and just keep one set of results.

I would like to see these tests be able to use `GreenboardUploader.upload` and drop all of this custom code. We can extend these tests to create a platform of `"sync-gateway-upload"` tests and then make sure `if platform == "sync-gateway" and sgw_version is not None:` accounts for this specific type of platform.

Sharing code is always preferable to not sharing code and using less code for this is very very helpful.

vipbhardwaj: The issue is every run creates a new document. This is the single most painful thing I find right now. I like the suggestion from Ben a lot though, that we can keep a document from a single run with an array of result Object for last 5 runs (can be last 10 or 7 whatever). 
For now, what is really in the need of, is the deletion of all the very very old-runs' documents and changing the old way to something of this sort. Which keeps the documents limited to a specific **run-type, with the recent-run-results** in it so that can be read by the UI.

As @torcolvin pointed out about the simple usage of `upload` method, that itself would need changing then, since the old implementation of it just goes and creates a document in a bucket with a strict format. 
FORMAT : is not the exact same for upgrade tests, hence the branching.
DOCUMENTS : cannot be new every time for each upgrade run, this would make
- fetching linearly poor overtime
- more format changing to filter out the recent `n` runs to be selected and shown by Greenboard.
- Just stupid clutter in the Database so that the _**code is non-smelly and nice**_ , which I really diagree with.

Honestly, I already have a way which highly intersects with @bbrks 's  advice.
I would already raise another commit with it but then it is extremely saddening to revert the days of work because PR approvals are a bit strict. 
Let's let me raise a commit related to what I am talking about, its fine if its frowned upon and has to be reverted. But it'll be clean overall and mostly have minor changes here and there, but overall I plan on changing this whole idea of : 
NEW document PER run behaviour of `GreenboardUploader.upload` , and refactor it as Ben said, plus some ideas. Mostly it'd be much cleaner and better than the current/legacy implementation since whenever.
**TLDR :** 
I want to keep it efficient (for UI) + scalable (for Database and the Backend in case more "platforms" have to be added.)

vipbhardwaj: @torcolvin @bbrks @borrrden Give me a go to start writing if this sounds good, else point me in the direction you guys find nice.
Next commit is about the comment below, coming soon. So the commit for THIS COMMENT will be post that, and post the approval for me to start writing code atleast (based on my above mental plan).

bbrks: One way to avoid the implement->pr->revert cycle is to come up with an agreed upon design or approach first. I know you've shared many design docs, etc. but there's never been dedicated time to review/iterate before you go and implement. We need to bake that into our processes a little bit, and intentionally slow down so we're not chasing our tails at review time.

I'm sure you're aware that working with >1 person involves >1 opinion on things - often directly conflicting - so everything is going to be a collaborative exercise to come up with reasonable compromises all-round. Really this is where the majority of the friction is coming from around these PRs. I do think most opinions are only loosely held or are suggestions not an actual ask, so figuring out what makes sense to intentionally not implement is of course the right thing to do to find that compromise.

I don't think the PRs are unnecessarily strict - we want to be writing maintainable code, otherwise we end up back in the same place as throwing out the legacy test framework because nobody understands it any more.

jenkins/pipelines/dev_e2e/android/android_tests.sh
```diff
@@ -44,4 +49,5 @@ echo "Run the tests"
 # To re-enable this, this script needs to become aware of the
 # serial number of the device, which is not currently passed
 #adb shell input keyevent KEYCODE_WAKEUP
-uv run pytest -v --no-header --maxfail=7 -W ignore::DeprecationWarning --config config.json --dataset-version $DATASET_VERSION
+uv run pytest -v --no-header --maxfail=7 -W ignore::DeprecationWarning --config config.json --dataset-version $DATASET_VERSION \
+    --junitxml="$GREENBOARD_RESULTS_DIR/junit_dev_e2e_android.xml"
```
torcolvin, line 53: As discussed in the shared channel with @borrrden on slack, let's figure out the simplest way to implement _just_ modifying junit files, and put this in a separate PR. Let's implement this for all test runs in one PR, and then we can worry about how to implement this for upgrade tests in a different PR.

1. add `junitxml=` argument to pytest.ini to say `junit_report.xml` which will always write this xml file to the cwd where pytest is launched
2. register that file via junit in jenkinsfiles

It might not be necessary to actually upload these xml files, because ultimately these artifacts are uploaded and stored on jenkins, and these don't need to be registered with `archiveArtifacts`

Can you create a PR which changes only this, and show an example of a mobile jenkins run that uses this output? I am also interested in how this looks for a pipeline job which wraps up multiple platforms http://jenkins.mobiledev.couchbase.com/view/QE/job/qe-java-pipeline/

After that PR looks OK, then we can think about how to extend the greenboard code in client/src/cbltest , but these are separate problems. For that case, we can override the names of the arguments to pytest with `-o junitxml`.

jenkins/pipelines/shared/config.sh
```diff
@@ -83,8 +83,26 @@ readonly QE_TESTS_DIR="$TESTS_DIR/QE"
 readonly QE_PIPELINES_DIR="$PIPELINES_DIR/QE"
 readonly AWS_ENVIRONMENT_DIR="$ENVIRONMENT_DIR/aws"
 
+# Per-Jenkins-build directory where every pytest invocation drops its
+# --junitxml output and the greenboard fixture drops its meta_*.json
+# sidecar. The aggregator (python -m cbltest.greenboard_upload) reads
+# this dir at the end of the build to produce one greenboard doc.
+# Default to $WORKSPACE on Jenkins agents; fall back to $PWD locally.
+readonly GREENBOARD_RESULTS_DIR="${WORKSPACE:-$PWD}/.greenboard-results"
```
torcolvin, line 91: I think you can avoid having any extra data in shell scripts to just picking up all files that are junit_*.xml in. the test files.

I would say anytime I see a shell script modified, especially if it is duplicated, I ask myself if there's a way to avoid writing this.

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -27,81 +55,66 @@ async def greenboard(cblpytest: CBLPyTest, pytestconfig: pytest.Config):
...
-                    os_name = test_server_info.device["systemName"]
-            if len(cblpytest.sync_gateways) > 0:
-                sgw_version = await cblpytest.sync_gateways[0].get_version()
-            uploader.upload(test_platform, os_name, library_version, sgw_version)
+        )
+        cbl_info(f"Wrote greenboard meta sidecar: {sidecar_path}")
     except Exception as e:
-        cbl_warning(f"Failed to upload results to Greenboard: {e}")
-    finally:
-        pytestconfig.pluginmanager.unregister(uploader)
```
torcolvin, line 122: I think this is intentionally unregistering the value - and in fact, I think there are several unintentional regressions.

The way I like to think about modifying code like this is to write tests against the existing behavior, so then the reviewer can see what code is changed. I wrote tests: https://github.com/couchbaselabs/couchbase-lite-tests/pull/418 for this code.

I think this part of this PR should be put on hold in favor of seeing how the junit output will be registered for the non upgrade case, so then I can understand how it should look/be modified for that case.

client/src/cbltest/greenboard_upload.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@click.command()
+@click.option(
+    "--config",
+    "config_path",
+    required=True,
+    type=click.Path(exists=True, dir_okay=False),
+    help="Path to the test config JSON (same one pytest was run with).",
+)
```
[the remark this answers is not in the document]
vipbhardwaj: Done

client/src/cbltest/greenboard_upload.py
```diff
@@ -0,0 +1,74 @@
...
+    parsed = _parse_config(config_path)
+    if (
+        parsed.greenboard_url is None
+        or parsed.greenboard_username is None
+        or parsed.greenboard_password is None
+    ):
+        cbl_info("No greenboard config in test config; skipping upload")
+        return
+
+    build_url = os.environ.get("BUILD_URL") or None
```
borrrden, line 54: I really prefer to see these sorts of things as command line options.  Making them environment variables makes them less easy to locate as the options are all just laid out at the top.

PR #415, opened 2026-05-19
Added fixes needed for QE Jenkins

sanjivanipatrax:
This PR adds in the other platforms to the Jenkinsfiles as well as fixes needed which were previously in `qe-jenkins` branch.

**Features added:**
- Added all platforms to the C Jenkinsfile (in `jenkins/pipelines/QE/c/Jenkinsfile`).
- The phone connected to mac-mini-CBL-Android1 node keeps intermittently disconnecting so changed it to use qe-mac-india node.
- Changed path in iOS Jenkinsfile to fix the git-lfs issue.
- Added post stage in iOS and .NET Jenkinsfile to get notified on regressions and successes (in `jenkins/pipelines/QE/dotnet/Jenkinsfile` and `jenkins/pipelines/QE/ios/Jenkinsfile`).

### DIFF
--- a/jenkins/pipelines/QE/c/Jenkinsfile
+++ b/jenkins/pipelines/QE/c/Jenkinsfile
@@ -14,6 +14,7 @@ pipeline {
                 script {
                     if (params.CBL_VERSION == '') { error "CBL_VERSION is required" }
                     if (params.SGW_VERSION == '') { error "SGW_VERSION is required" }
+
                     currentBuild.displayName = "C CBL:${params.CBL_VERSION} (#${currentBuild.number})"
                     currentBuild.description = "SGW: ${params.SGW_VERSION}"
                 }
@@ -23,10 +24,10 @@ pipeline {
             steps {
                 script {
                     def platforms = [
-                        'c_macos', 'c_linux_x86_64'
+                        'c_macos', 'c_ios', 'c_android', 'c_windows', 'c_linux_x86_64'
                     ]
                     def parallelBuilds = [:]
-                    for (p in platforms) {
+                    for(p in platforms) {
                         def platform = p
                         parallelBuilds[platform] = {
                             build job: 'prebuild-test-server',
@@ -52,7 +53,6 @@ pipeline {
                     environment {
                         PATH = "/opt/homebrew/opt/python@3.10/bin:/opt/homebrew/bin:/usr/local/bin:${env.PATH}"
                         TS_ARTIFACTS_DIR = 'c_macos'
-                        TS_DATASET_VERSION = "${params.TS_DATASET_VERSION}"
                     }
                     steps {
                         withCredentials([string(credentialsId: 'mobile-qe', variable: 'KEYCHAIN_PASSWORD')]) {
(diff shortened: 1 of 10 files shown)


jenkins/pipelines/QE/c/Jenkinsfile
```diff
@@ -43,110 +42,107 @@ pipeline {
...
                     }
                     environment {
-                        TS_DATASET_VERSION = "${params.TS_DATASET_VERSION}"
+                        PATH = "/opt/homebrew/opt/python@3.10/bin:/opt/homebrew/bin:/usr/local/bin:${env.PATH}"
+                        TS_ARTIFACTS_DIR = 'c_macos'
                     }
                     steps {
-                        echo "=== Run C Windows Tests"
+                        withCredentials([string(credentialsId: 'mobile-qe', variable: 'KEYCHAIN_PASSWORD')]) {
+                            sh 'security unlock-keychain -p ${KEYCHAIN_PASSWORD} ~/Library/Keychains/login.keychain-db'
```
borrrden: This is not needed for macOS, I think.  It's there in iOS to deal with code signing

jenkins/pipelines/QE/dotnet/Jenkinsfile
```diff
@@ -133,12 +133,12 @@ pipeline {
                     }
                 }
                 stage('Android') {
-                    agent { label 'mac-mini-CBL-Android1' }
+                    agent { label 'qe-mac-india' }
                     options {
-                        lock('mac-mini-CBL-Android1,R5CT20FYA7R')
+                        lock('qe-mac-india,51131XEKB7DBF1')
```
borrrden: Once again you've locked the Jenkins node.  Why?

borrrden (changes requested):

jenkins/pipelines/QE/c/Jenkinsfile
```diff
@@ -166,7 +162,33 @@ pipeline {
...
+                    options {
+                        lock('net-windows')
+                    }
+                    environment {
+                        TS_ARTIFACTS_DIR = 'c_windows'
+                    }
+                    steps {
+                        echo "=== Run C Windows Tests"
+                        timeout(time: 120, unit: 'MINUTES') {
+                            powershell "jenkins\\pipelines\\QE\\c\\run_test.ps1 -Version ${params.CBL_VERSION} -SgwVersion ${params.SGW_VERSION} -DatasetVersion ${params.TS_DATASET_VERSION}"
```
barkha06, line 164: The stage windows will fail, can you avoid modifying the part of the c-pipeline for windows, iOS and Android since I've made sure they run as expected. You can add the remaining platforms above or below these.

jenkins/pipelines/QE/dotnet/Jenkinsfile
sanjivanipatrax: Good catch - I was worried about other jobs interfering, but I just checked and the node has only one executor, so the lock is redundant. Removing it now

PR #417, opened 2026-05-25
feat: Adding context enhancement

vipbhardwaj:
Added the files that aid any model running via a **CLI or Installed** AI Application used to CRUD code in TDK.
These files makes it easier to break down and understand the context in TDK, since it is a horizontally scaled repo and has multiple files with very similar names.
This is not needed of course, but then it takes the same models multiple iterations to actually understand the files' intentions and understand where to precisely look at the quickest.

This also helps in saving tokens.

There is a root CLAUDE , COPILOT and FACTORY file. And then there are nested CLAUDE and FACTORY context files per directory to further define what the directory's files do and WHY THE DIRECTORY EXISTS.


docs/agent-context/build-test-matrix.md
```diff
@@ -0,0 +1,54 @@
...
+## Component-Specific Validation
+
+| Component | Command | Scope |
+|---|---|---|
+| `client/` tests | `uv run -- pytest --config client/tests/empty_config.json client/tests` | Framework tests used in GitHub workflow. |
+| `client/smoke_tests` | `cd client/smoke_tests && uv run pytest -x -v --config config_in.json` | Smoke checks requiring config file. |
+| `tests/dev_e2e` | `cd tests/dev_e2e && uv run pytest -x -v --config config.json` | Full developer E2E suite. |
+| `tests/QE` | `cd tests/QE && uv run pytest -x -v --config config.json` | Full QE suite. |
+| `tests/QE` SGW subset | `cd tests/QE && uv run pytest -x -v --config config.json -m sgw` | Marker-focused validation. |
+| AWS orchestrator lint/type | `uv run --group orchestrator-lint ty check` | Recommended when touching `environment/aws`. |
```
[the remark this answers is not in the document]
vipbhardwaj: Not needed

borrrden:
What are these new hooks for?  I don't plan on using them.

CLAUDE.md
```diff
@@ -0,0 +1,249 @@
...
+cd environment/aws && python stop_backend.py --topology topology_setup/topology.json
+```
+
+### Git Hooks
+
+All hooks are enforced automatically on commit via `.pre-commit-config.yaml`:
+
+- **Syntax & Style**: ruff (lint + import sorting), ruff-format, pyupgrade, ty check
+- **Merge Safety**: check-merge-conflict
+- **Commit Format**: Conventional Commits validation (`<type>[scope]: <description>`)
```
borrrden: Where did this convention come from?

borrrden:
Why are there both AGENTS.md and CLAUDE.md independently?  One should refer to the other because they both should be identical right?  Maintaining both seems like it will be a recipe for getting out of sync.

docs/agent-context/build-test-matrix.md
```diff
@@ -0,0 +1,54 @@
...
```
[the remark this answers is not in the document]
borrrden: This group does not exist anymore, copilot is correct.

environment/CLAUDE.md
```diff
@@ -0,0 +1,220 @@
...
+## Directory Structure
+
+```
+environment/
+├── aws/                        # AWS orchestrator (primary)
+│   ├── start_backend.py        # Main entry — provisions everything
+│   ├── stop_backend.py         # Teardown — destroys resources
+│   ├── main.tf                 # Terraform config (EC2, VPC, subnets)
+│   ├── download_tool.py        # Downloads cbbackupmgr and other tools
+│   ├── requirements.txt        # Python deps for orchestrator
```
borrrden: This file no longer exists

borrrden (changes requested):
I asked AI to review this, since it's meant for AI, and it's coming back with a lot of inconsistencies.  I suggest you do the same.

vipbhardwaj:
> Why are there both AGENTS.md and CLAUDE.md independently? One should refer to the other because they both should be identical right? Maintaining both seems like it will be a recipe for getting out of sync.

That is because AGENTS.md is what the droids from Factory look at. I don't think they'd get proper storable context from CLAUDE.mds? I haven't researched about this but I learnt 2 months back that AGENTS is for Factory. So I added contextual support for both. 

<hr/>

If there is a universal context file name that every AI tool understand, I'll switch, let me study a bit more for this.

environment/CLAUDE.md
vipbhardwaj: You pointed out 4, which one exactly?

CLAUDE.md
vipbhardwaj: I had been using it, its just a convention followed across open-source in general. I felt the commit messages were neater and shorter since we describe what's the commit generally about by the first word itself.

vipbhardwaj: I think we can keep it as a part of TDK, its not a big change, just asks for one-word key before the rest of the commit message.

vipbhardwaj:
So I looked into it and there were mostly adding SYMLINKS to CLAUDE.md and keeping context in AGENTS.md , or , still using the two files and adding `@ import`, but then I searched more and found that ONLY FOR claude-code, it falls back to AGENTS.md automatically in case of an absence of CLAUDE.md file. But when I confirmed this with claude-code, it denied.
I have gone with the `@ import` way since that still consolidates the data into a single file, avoiding double-file edits for the same change.

CLAUDE.md
borrrden: You can't just sneak it in without defining it somewhere and telling other committers about it.  This convention is new to me and the projects I look at it tend not to use it.

environment/CLAUDE.md
borrrden: When I make a comment like this, it's always the line at the very bottom of whatever Github shows

borrrden:
> So I looked into it and there were mostly adding SYMLINKS to CLAUDE.md and keeping context in AGENTS.md , or , still using the two files and adding `@ import`, but then I searched more and found that ONLY FOR claude-code, it falls back to AGENTS.md automatically in case of an absence of CLAUDE.md file. But when I confirmed this with claude-code, it denied. I have gone with the `@ import` way since that still consolidates the data into a single file, avoiding double-file edits for the same change.

Yes the import syntax is what I expect.  Symlinks don't work well on Windows.

environment/CLAUDE.md
vipbhardwaj: Oh, You can make it on just one line too

vipbhardwaj: Fixed

servers/AGENTS.md
```diff
@@ -0,0 +1,227 @@
...
+Per-platform test server implementations. Each one implements the same REST API (defined in [spec/api/api.yaml](../spec/api/api.yaml)) in the native language for its platform, so the Python `cbltest` framework can drive Couchbase Lite operations uniformly.
+
+## Scope
+
+You own everything under `servers/`:
+
+| Path | Language | Build System | Target Platforms | Transport |
+|---|---|---|---|---|
+| `c/` | C++ (C++17) | CMake 3.23+ | macOS, Linux, Windows, iOS, Android | HTTP (CivetWeb) |
+| `dotnet/` | C# (.NET 8) | MSBuild / `dotnet` CLI | Windows, macOS, iOS, Android | HTTP (Kestrel) |
```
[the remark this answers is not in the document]
borrrden: This is not entirely false.  .NET 10 is used to build, but for Windows .NET 8 is the deployment version.  It could be improved to distinguish between console and maui targets.

scripts/hooks/check_commit_msg.py
```diff
@@ -0,0 +1,199 @@
...
+BOGUS_PREFIX = re.compile(r"^[A-Za-z]+(\([^)]*\))?!?:\s*")
+
+
+def guidance(message: str) -> None:
+    """Print the non-interactive failure message (CI / no TTY)."""
+    types = " ".join(name for name, _ in TYPES)
+    lines = [
+        "ERROR: Commit message does not follow Conventional Commits format.",
+        "",
+        "  Expected: <type>[optional scope]: <description>",
```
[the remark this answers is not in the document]
vipbhardwaj: That is inlcuded in the optional scope and also clarified to the user while interactive changing, it does not need to be there for AI's context

.pre-commit-config.yaml
```diff
@@ -21,8 +21,8 @@ repos:
         entry: uv run ty check
       - id: conventional-commits
         name: Conventional Commits
-        language: script
-        entry: scripts/hooks/check-commit-msg.sh
+        language: system
+        entry: python3 scripts/hooks/check_commit_msg.py
```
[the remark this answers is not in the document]
vipbhardwaj: This repository works on python3 so it should be supported. python3 is the boilerplate for this repo

.factory/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "enabledPlugins": {
+    "core@factory-plugins": true
+  }
+}
```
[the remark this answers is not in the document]
vipbhardwaj: PR description is written by a human, by FACTORY he meant AGENTS.md files since FACTORY and COPILOT can refer to it, and he specifically wrote factory since FACTORY is what his company gives amongst other agents. Don't drill too much into it

client/AGENTS.md
```diff
@@ -0,0 +1,175 @@
...
+@register_body(TestServerRequestType.RESET, version=1)
+class PostResetRequestBodyV1(JSONSerializable): ...
+```
+
+### `TestServerRequestType` enum (in `requests.py`)
+
+18 values — keep in sync with the file:
+`ROOT`, `RESET`, `ALL_DOC_IDS`, `UPDATE_DB`, `START_REPLICATOR`, `REPLICATOR_STATUS`, `SNAPSHOT_DOCS`, `VERIFY_DOCS`, `PERFORM_MAINTENANCE`, `RUN_QUERY`, `GET_DOCUMENT`, `NEW_SESSION`, `LOG`, `START_LISTENER`, `STOP_LISTENER`, `START_MULTIPEER_REPLICATOR`, `STOP_MULTIPEER_REPLICATOR`, `MULTIPEER_REPLICATOR_STATUS`.
+
+(Note: `/stopReplicator` exists in the OpenAPI spec but the framework drives it through `Replicator.stop()` state transitions rather than a dedicated enum value.)
```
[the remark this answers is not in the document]
vipbhardwaj: I think that's clarified in the above enums so no stress.

vipbhardwaj:
@borrrden I have added the interactive Commit Convention thing so that it is even easier now to understand by other users.

scripts/setup-hooks.sh
borrrden, line 1: What are Windows users going to do about this? 😄

.pre-commit-config.yaml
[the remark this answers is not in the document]
borrrden: On Windows I'm not sure that's the case actually

scripts/setup-hooks.sh
vipbhardwaj: Removed commit convention 👍🏻

borrrden (approved):
I think some of this is going to fail on Windows but as long a

vipbhardwaj:
> I think some of this is going to fail on Windows but as long a

As long as?

borrrden:
> > I think some of this is going to fail on Windows but as long a
> 
> As long as?

This comment was for the previous iteration and no longer has any relevance.  I must have left it half typed and it sent it when I approved it...

borrrden (approved):

PR #418, opened 2026-05-27
Add tests for uploading documents to greenboard

torcolvin:
- Create a pydantic model to handle the data storage so future modifications are more clear.
- Allows for writing regression tests for upload simplification

This is a no-op modification to add regression tests for greenboard uploading so that modifications like https://github.com/couchbaselabs/couchbase-lite-tests/pull/414 can be more clear about what code they are or are not changing.

### DIFF
--- a/client/src/cbltest/greenboarduploader.py
+++ b/client/src/cbltest/greenboarduploader.py
@@ -9,9 +9,26 @@ from _pytest.reports import TestReport
 from couchbase.auth import PasswordAuthenticator
 from couchbase.cluster import Cluster
 from couchbase.options import ClusterOptions
+from pydantic import BaseModel, ConfigDict, Field
 
 from cbltest.api.syncgateway import CouchbaseVersion
 from cbltest.logging import cbl_info, cbl_warning
+
+
+class IntegrationTestRun(BaseModel):
+    """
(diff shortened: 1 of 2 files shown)


client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,28 @@
 from couchbase.auth import PasswordAuthenticator
 from couchbase.cluster import Cluster
 from couchbase.options import ClusterOptions
+from pydantic import BaseModel, ConfigDict, Field
 
 from cbltest.api.syncgateway import CouchbaseVersion
 from cbltest.logging import cbl_info, cbl_warning
 
 
+class IntegrationTestRun(BaseModel):
```
borrrden: Not exactly the name I'd choose.  We don't refer to these as integration tests anywhere.  If anything they are E2ETestRuns or TestRunResult, or something.

borrrden (changes requested):

client/src/cbltest/greenboarduploader.py
torcolvin: Yeah, I wanted to pick a name that didn't start with `Test` because that gets picked up as a test class by discovery.

borrrden (approved):

PR #419, opened 2026-06-01
Junit functional upload

vipbhardwaj:
[CBG-5403](https://jira.issues.couchbase.com/browse/CBG-5403)

[CBG-5403]: https://couchbasecloud.atlassian.net/browse/CBG-5403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


torcolvin:
Some of these fixtures are very similar to https://github.com/couchbaselabs/couchbase-lite-tests/pull/418, I'll see about @borrrden merging that PR so we can reuse them. It would be fine to move these to conftest.py for reuse in unit tests.

Do you have an example jenkins job that picks these up? You can fork this branch to just run in jenkins.

1. Modify a single Jenkinsfile so that we see the uploads in jenkins. (We might need to add a plugin to jenkins)
2. Change the test execution so that it runs only two tests (for speed of visualization), one failing test and one passing tests.
3. Link to the results

We want to make sure that we can see the results separately, which might need to set `junit_suite_name`, but until we see this in Jenkins, I'm not sure.

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -97,7 +98,39 @@ async def greenboard(cblpytest: CBLPyTest, pytestconfig: pytest.Config):
...
+                    cbl_info(
+                        "Greenboard: JUnit XML reports zero tests; skipping upload"
+                    )
+                    return
+                if junit_pass == 0:
+                    cbl_info(
+                        f"Greenboard: all tests failed (failCount={junit_fail}); "
+                        "skipping upload per policy"
+                    )
+                    return
```
torcolvin: This is a big block of code in an already big function - this can be a function that is `upload_from_junit_file(self, junit_output: pathlib.Path)`. If other arguments are needed like `test_platform`, `os_name`, etc, they can also be added to this function.

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -97,7 +98,39 @@ async def greenboard(cblpytest: CBLPyTest, pytestconfig: pytest.Config):
...
+                if junit_pass + junit_fail == 0:
+                    cbl_info(
+                        "Greenboard: JUnit XML reports zero tests; skipping upload"
+                    )
+                    return
+                if junit_pass == 0:
+                    cbl_info(
+                        f"Greenboard: all tests failed (failCount={junit_fail}); "
+                        "skipping upload per policy"
+                    )
```
torcolvin: If all tests fail, why do we not want to know about it?

client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,64 @@
...
+    counts from pytest's ``--junitxml`` output instead of in-process
+    hook-driven counters.
+    """
+    xml_path = Path(xml_path)
+    if not xml_path.is_file():
+        return None
+    try:
+        xml = JUnitXml.fromfile(str(xml_path))
+    except Exception as e:
+        cbl_warning(f"Failed to parse JUnit XML {xml_path}: {e}")
```
torcolvin: Do we actually not want to upload here? I'd be inclined to drop this exception handling.

If the upload code fails to run, and especially if this is at the end, I think we want to actually have this fail the jenkins job so we know this exists.

client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,64 @@
...
+    a JUnit XML file at ``xml_path``, or ``None`` if the file is missing or
+    can't be parsed.
+
+    ``passed = tests - failures - errors - skipped``; ``failed`` lumps
+    failures and errors together (the greenboard doc only carries a single
+    fail bucket). Used by the greenboard pytest fixture to derive upload
+    counts from pytest's ``--junitxml`` output instead of in-process
+    hook-driven counters.
+    """
+    xml_path = Path(xml_path)
```
torcolvin: This is already passed in as a `Path`, this doesn't need to be updated.

client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,64 @@
...
+    if not xml_path.is_file():
+        return None
```
torcolvin:
```suggestion
```

I would just let this fail and bubble an exception if this is not a file.

client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,64 @@
...
+    try:
+        xml = JUnitXml.fromfile(str(xml_path))
+    except Exception as e:
+        cbl_warning(f"Failed to parse JUnit XML {xml_path}: {e}")
+        return None
+
+    total_pass = 0
+    total_fail = 0
+    suites = list(xml) if isinstance(xml, JUnitXml) else [xml]
```
torcolvin:
```suggestion
```

Why would this be of different types?

client/src/cbltest/greenboarduploader.py
```diff
@@ -9,11 +9,64 @@
...
+    for suite in suites:
+        if not isinstance(suite, TestSuite):
+            continue
```
torcolvin: When is this not a test suite?

client/src/cbltest/plugins/greenboard_fixture.py
```diff
@@ -1,10 +1,11 @@
 import os
+from pathlib import Path
 
 import pytest
 import pytest_asyncio
```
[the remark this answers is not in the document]
vipbhardwaj: Fixed

client/tests/test_greenboarduploader.py
```diff
@@ -0,0 +1,595 @@
...
+        sync_gateways if sync_gateways is not None else []
+    )
+    return cblpytest
+
+
+def _make_pytestconfig(*, no_upload: bool = False) -> pytest.Config:
+    args = ["--config", "tests/empty_config.json"]
+    if no_upload:
+        args.append("--no-result-upload")
+    return pytest.Config.fromdictargs({}, args)
```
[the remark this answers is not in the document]
vipbhardwaj: Fixed

client/src/cbltest/greenboarduploader.py
vipbhardwaj: JUnit can return two types, but ours mostly would just be `JUnitXml` , hence fixed.

client/src/cbltest/plugins/greenboard_fixture.py
vipbhardwaj: I was originally inclined to show that up in greenboard, but that was denied. @borrrden , please discuss.

client/src/cbltest/greenboarduploader.py
torcolvin: If it returns a different type then I think if not isinstance(suite, TestSuite): is wrong. Do we want to do a type assertion and fail if it is not returning `JunitXml` ? What circumstances would it return a different type?

vipbhardwaj: Circumstances, I don't know honestly. What I read is that usually it returns either former or the latter : 
```
<testsuites>
  <testsuite name="..." tests="2" failures="0" ...>
    <testcase .../>
    <testcase .../>
  </testsuite>
</testsuites>
```
or 
```
<testsuite name="..." tests="2" failures="0" ...>
  <testcase .../>
  <testcase .../>
</testsuite>
```
Anyways, I read that its usually the `JUnitXml` in our case mostly, so I think it was dead code for us. Thanks for flagging that!

client/src/cbltest/greenboarduploader.py
vipbhardwaj: Actually almost never in our case, as our returned Type is always `JUnitXml` , therefore this is removed too.

client/src/cbltest/greenboarduploader.py
vipbhardwaj: So shall I remove the `try:catch` entirely or `raise e`, something else that you prefer?

torcolvin: The question is what happens in a pytest run after this? I think we'd want to flag this as a fatal error, so I'd be inclined to just raise this. If this happens then there was something pretty wrong with the infrastructure.

torcolvin (commented):
This can be in a separate PR, but I want a PR that adds to toplevel pytest.ini `tools.pytest.ini_options`

```
junit_xml = "junit_report.xml"
junit_output = "all"
junit_log_passing_tests = false # don't store output for passing tests
```

For the tests which self aggregate junit files, you can do `pytest -o junit_xml=custom_name.xml`

client/src/cbltest/greenboarduploader.py
vipbhardwaj: Done honestly

client/src/cbltest/greenboarduploader.py
```diff
@@ -63,6 +107,62 @@ def pytest_runtest_makereport(self, item: pytest.Item, call: pytest.CallInfo[Non
...
+            return
+
+        counts = count_from_junit_xml(junit_output)
+        assert counts is not None, f"Failed to parse JUnit XML at {junit_output}"
+        junit_pass, junit_fail = counts
+        if junit_pass + junit_fail == 0:
+            cbl_info(
+                f"Greenboard: JUnit XML at {junit_output} reports zero tests; "
+                "skipping upload"
+            )
```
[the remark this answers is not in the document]
vipbhardwaj: counts / assert is necessary for the ty check to pass since the function can return a list or NoneType

client/src/cbltest/greenboarduploader.py
```diff
@@ -80,6 +108,60 @@ def pytest_runtest_makereport(self, item: pytest.Item, call: pytest.CallInfo[Non
         elif report.failed:
             self.__fail_count += 1
 
+    def upload_from_junit_file(
+        self,
+        junit_output: Path,
+        platform: str,
+        os_name: str,
+        version: str,
+        sgw_version: CouchbaseVersion | None,
```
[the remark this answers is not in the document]
vipbhardwaj: I am testing those, by running the tests and checking doc uploads

client/src/cbltest/plugins/greenboard_fixture.py
borrrden: I feel like I've been consistent in saying that the thing I don't want in greenboard is not a run where all tests failed, but a run in which all tests _failed to run_ (i.e. had a setup failure or something wrong with the test harness, and thus no worthwhile results are generated).

client/src/cbltest/greenboarduploader.py
```diff
@@ -143,8 +238,8 @@ def upload(
                 build=parsed_build,
                 version=parsed_version,
                 sgwVersion=sgw_version_str,
-                failCount=self.__fail_count,
-                passCount=self.__pass_count,
+                failCount=resolved_fail,
```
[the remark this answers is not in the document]
borrrden: Actually it would probably be better to know about this quickly.  If the results aren't published to greenboard because greenboard is down, that's a problem worth flagging by failing the run.

scripts/hooks/check-commit-msg.sh
borrrden, line 1: This file has snuck in again?  It is not related to Junit.

borrrden (approved):
I think this is ok as long as the stray file for checking commit message is removed

scripts/hooks/check-commit-msg.sh
vipbhardwaj: Thanks for pointing it out.
Removed.

vipbhardwaj:
Since [this](https://github.com/couchbaselabs/couchbase-lite-tests/pull/419#pullrequestreview-4402459656) can be in a separate PR, and Tor is in a recovery for a while, I will merge this PR for now, based on resolving all of his comments and Jim's PR approval as well.

I will work on that separate PR related suggestion and will implement it as a part of Upgrade Test related Junit implementation, SINCE THIS PR ONLY CONCERNS WITH FUNCTIONAL TESTS, hence no AGGREGATION of pytest files.

PR #420, opened 2026-06-03
Add support for TLS on JAK testserver

barkha06:
* Created a KeyStoreService class that defines getTLSIdentity. 
* Both Java and Android have different TLSIdentity method definitions so they each implement a subclass of KeyStoreService
* The getTLSIdentity method stores the generated identity data in the keystore and returns an instance of TLSIdentity class. If the idenitity is null and disableTLS is false, the method returns an existing identity from the keystore via the alias.
* Port is dynamically assigned by OS to avoid any address in use errors
* Added timout between test steps to avoid server disconnected errors

### DIFF
--- a/tests/QE/test_peer_to_peer.py
+++ b/tests/QE/test_peer_to_peer.py
@@ -76,15 +76,15 @@ class TestPeerToPeer(CBLTestClass):
         all_dbs = [dbs[0] for dbs in all_devices_dbs]
 
         self.mark_test_step("Start listener on Device-1")
-        listener = Listener(all_dbs[0], ["_default._default"], 59840)
+        listener = Listener(all_dbs[0], ["_default._default"])
         listener.set_identity()
         await listener.start()
-        await asyncio.sleep(0.3)
         self.mark_test_step(f"listener started at {listener.port}")
-        port = listener.port if listener.port is not None else 59840
         self.mark_test_step(f"Add {num_of_docs} docs to Device-2")
         documents = await self._testserver_crud(all_dbs[1], num_of_docs)
-        endpoint = cblpytest.test_servers[0].replication_url("db1", port, tls=True)
+        endpoint = cblpytest.test_servers[0].replication_url(
+            "db1", listener.port, tls=True
+        )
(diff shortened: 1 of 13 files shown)


tests/QE/test_peer_to_peer.py
```diff
@@ -599,10 +607,10 @@ async def test_peer_to_peer_with_server_down(
         await replicator1.start()
 
         async def stop_restart_task():
+            port = listener1.port
             await listener1.stop()
-            listener2 = Listener(
-                all_dbs[0], ["_default._default"], 59840, identity=None
-            )
+            await asyncio.sleep(30)
```
borrrden: This is an excessively long sleep.  What is it for?  Please add a comment, the function name does not imply that this needs any wait time.

borrrden (changes requested):

tests/QE/test_peer_to_peer.py
barkha06: This was added because java testserver on linux OS requires longer time to free up a port, the test had failed serveral times due to address in use errors.

barkha06: There's also a situation where the port never frees up - https://jira.issues.couchbase.com/browse/CBL-8356.

barkha06: I've added a comment justifying the same

borrrden: Where is the comment?

borrrden: Nevermind a refresh made it show up

borrrden: The answer to "Java Linux sometimes misbehaves" should not be "add a 30 second unconditional test delay to every platform"

borrrden (changes requested):
At least gate the massive wait in stop restart to only happen for Java

pasin:
@barkha06 can you check Jim's comment about "At least gate the massive wait in stop restart to only happen for Java"?

barkha06:
Added a check for CBL Java for the delay

PR #422, opened 2026-06-10
Added JobURL basics + gitignored XML

vipbhardwaj:
[CBG-5401](https://jira.issues.couchbase.com/browse/CBG-5401)

[CBG-5401]: https://couchbasecloud.atlassian.net/browse/CBG-5401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/cbltest/greenboarduploader.py
+++ b/client/src/cbltest/greenboarduploader.py
@@ -53,6 +53,28 @@ def count_from_junit_xml(xml_path: Path) -> tuple[int, int, int]:
     return total_pass, total_fail, total_error
 
 
+def resolve_job_url() -> str:
+    """Return the Jenkins build URL for the current run, or ``"local"`` off-CI.
+
+    Jenkins exports ``BUILD_URL`` automatically for every build — the URL
+    of the specific run's status page (e.g.
+    ``https://jenkins.example.com/job/foo/123/``). That's exactly the
+    target greenboard's bar/matrix graphs should deep-link to, so we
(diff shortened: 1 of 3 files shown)


client/src/cbltest/greenboarduploader.py
```diff
@@ -53,6 +53,28 @@ def count_from_junit_xml(xml_path: Path) -> tuple[int, int, int]:
...
+    render them without dangling at a dead URL.
+
+    Resolved at upload time (not at uploader construction) because the
+    upgrade-batch path runs from a separate Python process invoked by
+    the wrapper shell script — the same Jenkins env is still present
+    there since it's the same job, and reading env on each call keeps
+    a single source of truth without threading the value through
+    constructors.
+    """
+    return os.environ.get("BUILD_URL") or "local"
```
borrrden, line 75: I'm not sure that this should return anything if run outside CI.  At least it shouldn't show up in greenboard in my opinion, but I don't know if you've considered that or not.

borrrden (approved):

client/src/cbltest/greenboarduploader.py
vipbhardwaj: I see, so a `return os.environ.get("BUILD_URL")` would be enough? I wasn't sure what would it specify for the url param for a local run since they get uploaded too.

PR #424, opened 2026-06-22
Added retry loop

vipbhardwaj:
[CBG-5486](https://jira.issues.couchbase.com/browse/CBG-5486)

[CBG-5486]: https://couchbasecloud.atlassian.net/browse/CBG-5486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/cbltest/api/syncgateway.py
+++ b/client/src/cbltest/api/syncgateway.py
@@ -970,6 +970,31 @@ class _SyncGatewayBase:
             assert isinstance(resp, dict)
             return AllDocumentsResponse(cast(dict, resp))
 
+    async def wait_for_all_documents(
+        self,
+        db_name: str,
+        min_count: int,
+        scope: str = "_default",
+        collection: str = "_default",
+        max_retries: int = 30,
+        retry_delay: int = 1,
+    ) -> AllDocumentsResponse:
+        """
+        Poll _all_docs until at least min_count docs are present, or time out.
+
+        Docs that arrive via an asynchronous path (SDK import, re-import after an
+        SGW restart, cross-node/ISGR propagation) are not guaranteed to be visible
+        in a single read. Returns the last response either way so the caller can
+        assert and report how far it got.
+        """
+        all_docs = await self.get_all_documents(db_name, scope, collection)
+        for _ in range(max_retries):
+            if len(all_docs.rows) >= min_count:
+                break
+            await asyncio.sleep(retry_delay)
+            all_docs = await self.get_all_documents(db_name, scope, collection)
+        return all_docs
+
     async def get_changes(
         self,
         db_name: str,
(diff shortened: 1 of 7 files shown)


tests/QE/test_server_setup.py
```diff
@@ -47,27 +46,15 @@ async def test_sgw_server_alternative_address(self, cblpytest: CBLPyTest) -> Non
...
-        expvars_dict = cast(dict, expvars)
-        import_count = (
-            expvars_dict.get("syncgateway", {})
-            .get("per_db", {})
-            .get(sg_db, {})
-            .get("shared_bucket_import", {})
-            .get("import_count", 0)
-        )
+        import_count = await sg.wait_for_import_count(sg_db, num_docs)
         assert import_count != 0, f"Expected import_count > 0, got {import_count}"
```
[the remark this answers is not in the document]
vipbhardwaj: We just need to make sure via expvars endpoint that imports have been underway or now, and this function implements a retry loop for it to be verified as the old code was running into a race saying NO DOCS imported as it reached the snippet earlier compared to the speed of API being able to report the import of docs.

This change is also okay to check for just `>0` since we do not need to assert for all `num_docs` be there, the test focuses on the import working and having it checked specifically via `/_expvars` . Hence `num_docs` is going to take just more time to reach for all docs and isn't adding any further help to the test

client/src/cbltest/api/syncgateway.py
```diff
@@ -970,6 +970,31 @@ async def get_all_documents(
             assert isinstance(resp, dict)
             return AllDocumentsResponse(cast(dict, resp))
 
+    async def wait_for_all_documents(
```
borrrden, line 981: I'd like to see functions that retry using the python library tenacity going forward

borrrden (approved):

client/src/cbltest/api/syncgateway.py
vipbhardwaj: Will do right now, AND, going forward

tests/QE/test_server_setup.py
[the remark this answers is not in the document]
vipbhardwaj: Done

client/src/cbltest/api/syncgateway.py
vipbhardwaj: Done

client/src/cbltest/api/syncgateway.py
```diff
@@ -2049,6 +2078,33 @@ async def wait_for_db_up(
         # Wait for the node to settle down after coming online
         await asyncio.sleep(settle_online)
 
+    @tenacity.retry(
+        wait=tenacity.wait_random_exponential(multiplier=1, max=10),
```
borrrden: If the polling is about 10 seconds, is there a reason to retry with an initial value of 1 second?  This sounds like a case for a fixed interval rather than a random exponential one.

borrrden (changes requested):
I have the same comment on both retry attributes.  I don't think exponential backoff is appropriate here if you know there is another attempt window at fixed intervals of 10 seconds.  This would mean that if you are at the beginning of it, you'd try again at approximately 1 with a guaranteed fail, 2 with a guaranteed fail, 4 with a guaranteed fail, and then 8 might succeed for a total wait time of 15 seconds.  It seems that if what you are waiting for is the 10 seconds for another import scan, just wait for 10 seconds each time (or 11)

borrrden (approved):

PR #430, opened 2026-07-02
Added options to run only specific platforms and skip prebuild for QE pipelines

sanjivanipatrax:
This PR includes changes for running the tests for specific CBL builds and option to skip prebuild job.

**Features added:**
- Per-platform checkboxes on the multi-platform QE Jenkinsfiles (dotnet, c, java) so only the selected platforms run tests.
- A `RUN_PREBUILD` checkbox on all five QE pipelines (dotnet, c, java, ios, android) - unchecked skips prebuild and reuses existing artifacts.
- Guarded each stage with `when { beforeAgent true; ... }` so skipped platforms don't wait for executors.

Verified against a duplicated job pointed at this branch.

### DIFF
--- a/jenkins/pipelines/QE/android/Jenkinsfile
+++ b/jenkins/pipelines/QE/android/Jenkinsfile
@@ -4,6 +4,7 @@ pipeline {
         string(name: 'CBL_VERSION', defaultValue: '3.3.0', description: 'Couchbase Lite Version (build number will be fetched if not provided)')
         string(name: 'SGW_VERSION', defaultValue: '3.3.0', description: "The version of Sync Gateway to download")
         string(name: 'TS_DATASET_VERSION', defaultValue: '4.0', description: 'The version of the dataset to use on the test servers')
+        booleanParam(name: 'RUN_PREBUILD', defaultValue: true, description: 'Prebuild the test server. Uncheck to skip prebuild entirely and reuse existing artifacts.')
     }
     stages {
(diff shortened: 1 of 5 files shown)


jenkins/pipelines/QE/android/Jenkinsfile
```diff
@@ -4,6 +4,7 @@ pipeline {
         string(name: 'CBL_VERSION', defaultValue: '3.3.0', description: 'Couchbase Lite Version (build number will be fetched if not provided)')
         string(name: 'SGW_VERSION', defaultValue: '3.3.0', description: "The version of Sync Gateway to download")
         string(name: 'TS_DATASET_VERSION', defaultValue: '4.0', description: 'The version of the dataset to use on the test servers')
+        booleanParam(name: 'RUN_PREBUILD', defaultValue: true, description: 'Prebuild the test server. Uncheck to skip prebuild entirely and reuse existing artifacts.')
```
borrrden, line 7: This is already true today.  The prebuild job is a no op if the test server already exists.

borrrden (approved):

jenkins/pipelines/QE/android/Jenkinsfile
sanjivanipatrax: Yes, but this is tighter - it avoids triggering the prebuild job at all, and saves queuing for the executor too.

PR #432, opened 2026-07-03
Added parallel sg_collect for all nodes from config

vipbhardwaj:
[CBG-5487](https://jira.issues.couchbase.com/browse/CBG-5487)

This PR adds sg_collect post test suite run. Very small changes, created the method to run curl commands on all nodes for sg_collect to dump the logs into SUPPORTAL.

### KEYS : 
1. sg collect is triggered in `tests/QE/conftest.py` as a session fixture for a suite if one of the tests in it was marked with `sgw` pytest marker this PR. I plan on adding this into SGW for dev_e2e LATER in a different ticket, once the corrections and comments on this are finalized for sure.
2. They are run on all nodes in parallel because QE uses 3 SGs and sequential led to about 17mins of extra test runtime plus what we already face.
3. They dump the logs into the Jenkins job itself as its build artifacts. Hence they stay temporarily compared to the previous approach.
4. Kept the logic in a new file `environment/aws/sg_collect.py` which is very much similar to `stop_backend.py`.

I have tried to keep the code as concisie and clean, and as parallel and fast, AND AS THREAD-SAFE + RACE-SAFE as possible. 
Please give recommendations for improvements in this.

[CBG-5487]: https://couchbasecloud.atlassian.net/browse/CBG-5487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #436, opened 2026-07-09
CBG-5487 add an option to run sgcollect for failed runs

torcolvin:
CBG-5487 add an option to run sgcollect for failed runs

If `--sgcollect-on-test-failure` is set, then sgcollects are run at the end of pytest and output to the cwd. This only works in a caddy setup, which is OK for jenkins.

- Raise a cbl_error but do not fail the run if sgcollect output fails. Hopefully this is enough to raise that there is a problem. I'm open to failing the test harness as well.
- I wanted to raise an error without a noisy traceback, so added an option to avoid printing the stack. The stack if this fails is strictly unhelpful. Is there a better way to flag something really important? I am happy to also raise an traceback in the traditional style.
- Moved sgcollect options to use pydantic so I stopped mistyping all the options.
- It sucks that sgcollect doesn't output the name of the zipfile that it finds, so I guess based on new zipfiles written.

This is some tweaks on https://github.com/couchbaselabs/couchbase-lite-tests/pull/432

A few questions for @borrrden:

- Do you think this is useful to turn on always or have a flag for in jenkins to rerun with this flag? I can see this not being useful for a large collection of tests, but if there's a replication failure then it is.
- Did I contaminate cblpytest too much? I tried to keep noise out of there.


client/src/cbltest/plugins/cblpytest_fixture.py
```diff
@@ -41,7 +44,12 @@ async def cblpytest(request: pytest.FixtureRequest):
         config, log_level, test_props, dataset_version=dataset_version
     )
     yield cblpytest
-    await cblpytest.close()
+
+    try:
+        if sgcollect_on_test_failure and request.session.testsfailed:
```
vipbhardwaj: This restricts sg_collect to failed tests, which is not what the intention is since we want to at times collect logs for passing runs as well.
I think keeping a flag is fine but not a condition for ONLY_FAILING_TESTS.

jenkins/pipelines/shared/config.sh
```diff
@@ -31,6 +31,13 @@ function move_artifacts() {
     # Include the JUnit XML so each platform's results are preserved per
     # artifacts dir in a multi-pipeline (matrix) build.
     mv "$src_dir/junit_result.xml" "$dst_dir/junit_result.xml" || true
+    # SGW diagnostics downloaded by run_sgcollects() (via --sgcollect-on-test-failure)
```
vipbhardwaj, line 34: I think in #432 I had already added a method / way in which the destination directory is cleared before the new artifacts move to it

vipbhardwaj (changes requested):
I have a few points to get your opinion on, but in general it looks fine, although I feel this PR is in general just a re-ordering of code to make it Object Oriented.

jenkins/pipelines/shared/config.sh
torcolvin: This PR supersedes https://github.com/couchbaselabs/couchbase-lite-tests/pull/432 since that PR did not address the following comments here:

https://github.com/couchbaselabs/couchbase-lite-tests/pull/432#discussion_r3530014057

> I was thinking to add a pytest argument --run-sgcollect-on-failure. Then, you can add this call before cblpytest.close()

That PR did not add an option.

> [couchbase-lite-tests/client/src/cbltest/plugins/cblpytest_fixture.py](https://github.com/couchbaselabs/couchbase-lite-tests/blob/9c2bf0a49f8a4e88fedbe0a7f3663336df8b9fc0/client/src/cbltest/plugins/cblpytest_fixture.py#L38)> 
Line 38 in [9c2bf0a](https://github.com/couchbaselabs/couchbase-lite-tests/commit/9c2bf0a49f8a4e88fedbe0a7f3663336df8b9fc0)
> await cblpytest.close() 
> Then you can run this without using the QE script.

https://github.com/couchbaselabs/couchbase-lite-tests/pull/432 persisted in duplicating the work already existing in sgcollect.py

torcolvin:
> I have a few points to get your opinion on, but in general it looks fine, although I feel this PR is in general just a re-ordering of code to make it Object Oriented.

I'm not sure I understand that? I implemented what I requested in CBG-5487 - the ability for tests to output sgcollect to the current directories. Because I wanted to change the arguments to pull unredacted logs and run the code in a parallel without writing any new code to coordinate with sgcollect other than what was in syncgateway.py

client/src/cbltest/plugins/cblpytest_fixture.py
torcolvin: I think this is the behavior we want?

What is the use case for collecting sgcollect when we know the tests are passing?

jenkins/pipelines/shared/config.sh
vipbhardwaj: I am not sure why do we need to have the flag, it was decided to in general collect logs for even passing tests since recently we had an example where the BUG requested logs for a passing run as well. In order to help distinguish behavior and increase precision for Bug Fix

client/src/cbltest/plugins/cblpytest_fixture.py
vipbhardwaj: > recently we had an example where the BUG requested logs for a passing run as well. In order to help distinguish behavior and increase precision for Bug Fix

Hence I think its better we revert the flag? or keep it anyways, doesn't harm in any way actually.

vipbhardwaj:
I think we can merge this, there are a couple of points I would like to addrerss : 
1. [CBG-5487](https://jira.issues.couchbase.com/browse/CBG-5487) had old description from when I started and then team chipped in reviews which changed code but the ticket is not yet updated, I have updated it now and hence has what #432 actually implements post-changes
2. The flag added in this PR is **not harmful** but also according to me not necessary, we can keep it since it is just an option. But since it would not overload the space on the host node (because Jenkins jobs are configured to only keep runs/jobs till a certain `number && time` (whichever comes first)), therefore we can actually just collect logs for all runs, because that's harmless too.

[CBG-5487]: https://couchbasecloud.atlassian.net/browse/CBG-5487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

environment/local/CLAUDE.md
```diff
@@ -0,0 +1,115 @@
+# Local Test Environment
+
+Run the full TDK stack (Sync Gateway + test server) on your own machine, against either **rosmar**
+(in-memory, no Couchbase Server needed) or **cbs** (real Couchbase Server). This is the fastest
+inner loop for debugging a single failing test — no AWS provisioning, no Terraform, full local logs.
```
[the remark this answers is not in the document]
torcolvin: this didn't belong in this PR, removed.

jenkins/pipelines/shared/config.sh
vipbhardwaj, line 30: We should do a removal of stale `http_log` dir in the `dest_dir` as well, since i am seeing in the recent runs that there are stales of http_log , from old runs with old IPs.

Image

[URL](http://jenkins.mobiledev.couchbase.com/view/QE/job/sgw-functional/169/artifact/tests/QE/sgw/http_log/) to the same.

vipbhardwaj (commented):
Just a small but important detail

jenkins/pipelines/shared/config.sh
borrrden: I actually noticed this as well....this is important enough that I'll just commit a one line fix to main directly.  I don't think anyone needs to bother looking at the one line in an official PR and being like "year that's a good one line"

I think all we need to do it just `rm -rf $dst_dir` before line 28 and no more stale artifacts ever again.

torcolvin: I am not 100% sure that I understand, but I didn't want this comment to get lost so I created a separate PR https://github.com/couchbaselabs/couchbase-lite-tests/pull/448

PR #442, opened 2026-07-13
CBG-5579 avoid 503 db under maintainence flakes

torcolvin:
CBG-5579 avoid 503 db under maintenance flakes

**Direct fixes:**

- Add call for wait database after restoring from cbbackupmgr rather than relying on a 2s sleep and a config polling window for whether the db config will exist or not. That is in `tests/shared/upgrade_test_helpers.py`. Everything else is helper code that is in `client` code to make errors surface to developers faster.

**Improvements:**

Because of the subtlety of this bug, I desired to fix this in a more robust and comprehensive way. Create a pattern of `SyncGatewayCluster` to represent multiple Sync Gateway clusters, since this is a common pattern for needing to poll for db online.

- Turn the polling for waiting for database online to use `/_all_dbs?verbose=true` rather than `GET /db`. This drops the requirement that we have to handle a 404/403 error when the database doesn't exist and we can expect this function to not throw any exceptions.
- create assert_retry function that makes reporting failures clear and use for updated functions. Used some real pytest magic for this to occur but it is very hidden. use __tracebackhide__ https://docs.pytest.org/en/stable/example/simple.html#writing-well-integrated-assertion-helpers and tenacity retry wrappers.
- Remove a wait_for_db_up call immediately after put_database if there is only one node, PUT /db is already synchronous.
- Drop settle_online parameter. If the database is reported as online, it should work and if it doesn't, that's a SG bug. This is an aside from the fixes.
- Drop the retry window to 1s to speed up test runtime, polling GET /db/ is cheap.

When calling TestReplicationUpgrade, there were flakes in:

```
>               raise CblSyncGatewayBadResponseError(
                    resp.status, f"{method} {path} returned {resp.status}: {data}"
                )
E               cbltest.api.error.CblSyncGatewayBadResponseError: get /upgrade._default._default/conflict_6 returned 503: {
E                 "error": "Service Unavailable",
E                 "reason": "DB is currently under maintenance"
E               }
```

This occurs because of the way the database restore works. cbbackupmgr contains all the config, checkpoints (_sync docs) from an older database with different vBucket UUIDs and can be slow to restore. Additionally, CBG-5539 started in Sync Gateway 4.1 which means that many operations are slow once a bucket has been deleted by couchbase server due to being stuck polling for the deleted bucket by cluster compat manager. Each of these bugs is worth fixing in Sync Gateway, but none are worth holding up any versions of Sync Gateway testing.

Pattern used that results in the 503 operations. Since there is only a short sleep time and it doesn't wait for the database to come online:

```
  test_case.mark_test_step("Delete Sync Gateway 'upgrade' database if exists")
    # delete_database silently swallows 403 internally (config-managed DBs)
    # and retries 500s; no need to wrap further.
    await cblpytest.sync_gateways[0].delete_database("upgrade")

    test_case.mark_test_step("Restore Couchbase Server Bucket using `upgrade` dataset")
    cbs: CouchbaseServer = cblpytest.couchbase_servers[0]
    cbs.drop_bucket("upgrade")
    # reset_expired_ttl restores the delta-sync old-revision backup bodies
    # (`_sync:rev:*`) so SGW can delta against a legacy ancestor rev.
    cbs.restore_bucket(
        "upgrade",
        tools_path(),
        dataset_path,
        "upgrade",
        reset_expired_ttl=reset_expired_ttl,
    )

    test_case.mark_test_step("Wait 2s to ensure SG picks up the restored database.")
    await asyncio.sleep(2)
```    

Tests locally and with https://jenkins.sgwdev.com/job/Couchbase%20Lite%20E2E/30/ . I also ran all QE tests (though not with AWS).


client/src/cbltest/api/syncgateway.py
```diff
@@ -550,30 +551,53 @@ class SyncGatewayStatusResponse(BaseModel):
...
 
-    @property
-    def db_name(self) -> str:
-        """Gets the database name"""
-        return self.__db_name
+    ONLINE = "Online"
+    OFFLINE = "Offline"
+    STARTING = "Starting"
+    STOPPING = "Stopping"
+    RESYNCING = "Resyncing"
```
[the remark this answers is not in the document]
torcolvin: There isn't an unknown state.

client/src/cbltest/utils.py
```diff
@@ -3,13 +3,48 @@
...
+import tenacity.asyncio
 
 from .api.error import CblTimeoutError
 
 T = TypeVar("T")
 
+# Hide tenacity's retry-loop frames so failures show the actual assertion, not
+# AsyncRetrying plumbing.
+tenacity.asyncio.__dict__["__tracebackhide__"] = True
+tenacity._utils.__dict__["__tracebackhide__"] = True
```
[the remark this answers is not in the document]
torcolvin: This comment is not correct. I'm not very worried about fragility of this, this just shortens tracebacks when we can.

tests/QE/test_rolling_upgrade_sgw.py
```diff
@@ -87,8 +87,6 @@ async def test_rolling_upgrade_sgw_cluster(
             except CblSyncGatewayBadResponseError as e:
                 if e.code != 412:
                     raise e
-        for sg in sg_nodes:
```
vipbhardwaj, line 90: Shouldn't we be adding `wait_for_db_online` for the removal of this as well?

tests/QE/test_server_setup.py
```diff
@@ -26,15 +26,14 @@ async def test_sgw_server_alternative_address(self, cblpytest: CBLPyTest) -> Non
             "import_docs": True,
         }
         await sg.put_database(sg_db, PutDatabasePayload(db_config))
-        await sg.wait_for_db_up(sg_db)
```
vipbhardwaj, line 29: I see its not been added after every `put_database`, but only selective, I am not able to wrap my head around the reason to that 🤔

tests/QE/test_ttl.py
```diff
@@ -36,7 +36,6 @@ async def test_document_expiry_unix_timestamp(
         }
         db_payload = PutDatabasePayload(db_config)
         await sg.put_database(sg_db, db_payload)
```
vipbhardwaj, line 38: Same goes for these and other  as well

tests/QE/test_xattrs.py
```diff
@@ -39,7 +39,6 @@ async def test_offline_processing_of_external_updates(
         }
         db_payload = PutDatabasePayload(db_config)
         await sg.put_database(sg_db, db_payload)
```
vipbhardwaj, line 41: All in `test_xattrs.py`

tests/QE/test_replication_upgrade_delta_sync.py
```diff
@@ -89,7 +89,6 @@ async def _prepare_sg_with_delta_sync(self, cblpytest: CBLPyTest) -> None:
             except CblSyncGatewayBadResponseError as e2:
```
vipbhardwaj, line 89: Here

tests/QE/test_db_gone.py
```diff
@@ -90,7 +90,6 @@ async def test_db_gone_on_bucket_deletion(
         }
         db_payload = PutDatabasePayload(db_config)
         await sg.put_database(sg_db, db_payload)
```
vipbhardwaj, line 92: This whole file too

tests/QE/test_large_doc_workloads.py
```diff
@@ -58,7 +58,6 @@ async def test_oversized_attachment_push(self, cblpytest: CBLPyTest) -> None:
         }
         db_payload = PutDatabasePayload(db_config)
         await sg.put_database(sg_db, db_payload)
```
vipbhardwaj, line 60: Same here

tests/QE/test_log_redaction.py
```diff
@@ -109,7 +109,6 @@ async def test_log_redaction_partial(
         }
         db_payload = PutDatabasePayload(db_config)
         await sg.put_database(sg_db, db_payload)
```
vipbhardwaj, line 111: Same

tests/QE/test_multiple_servers.py
```diff
@@ -75,7 +80,6 @@ async def _setup_database_and_user(
         "scopes": {"_default": {"collections": {"_default": {}}}},
     }
     await sg.put_database(sg_db, PutDatabasePayload(db_config))
```
vipbhardwaj, line 78: Added in this file optionally, I get that there might be some logic to it ofcourse but I feel that I had added waits after every db reset/put for avoiding races since I was running into them back in the day

tests/QE/test_server_setup.py
torcolvin: put_database is a synchronous operation. If there's a single Sync Gateway, a wait is not required.

tests/QE/test_rolling_upgrade_sgw.py
torcolvin: put_database is a synchronous operation. If there's a single Sync Gateway, a wait is not required. see other cases below.

vipbhardwaj: Oh I wasn't aware of that

vipbhardwaj: Wait so for multiple sync gateways we need it because every put_database is returning a session resp and we want to wait till the LAST response before moving forward?

borrrden (approved):
Ok my concerns about the scattering of information inside test signature is gone now

PR #443, opened 2026-07-13
CBL-8442: Conditional runtime permissions grants for Android Test Server

borrrden:
.NET and C don't support multipeer at all and so don't need any.  Java Android needs it based on the API level of the device


vipbhardwaj (approved):
LGTM

PR #444, opened 2026-07-16
Increase line length to 120

torcolvin:
This is only a whitespace change, increased line length in pyproject.tom and ran ruff format.

I'm not set at all on doing this, I just find some assertions easier to read on 1 line.


borrrden:
The thing about merging PRs that touch pretty much every file is that you are almost guaranteed to create merge conflicts with existing PRs that may or may not be a struggle already so I think this needs to get treated as a sort of barrier wait.  Things already in line need to get resolved first and this PR will absorb the rebase pains.

torcolvin:
I think dropping this PR is fine, I suspect there is never a good time to merge this.

PR #445, opened 2026-07-17
Support multiple Sync Gateway in a CouchbaseCloud

torcolvin:
Having CouchbaseCloud.sync_gateway was the wrong abstraction since there can be multiple Sync Gateways.

Use SyncGatewayCluster.wait_for_db_online to wait for the database on all nodes before continuing.

Stubbed some round_robin_node and random_node functions but they aren't used as part of this API. This is useful for future multiple node coordination tasks, like resync tests. Added here for some demonstration.

Having some abstractions is useful for the creating a database tasks. The workflow is:

1. POST /db on node1 - this operation is synchronous to come online
2. node2 (and others) are on a 10s polling loop to look for a documents in each bucket. When they find a new document or updated document, the database on node2 is uploaded.

### DIFF
--- a/client/src/cbltest/__init__.py
+++ b/client/src/cbltest/__init__.py
@@ -215,8 +215,8 @@ class CBLPyTest:
 
         if len(self.couchbase_servers) == 0:
             if self.sync_gateways[0].using_rosmar:
-                return CouchbaseCloud(self.sync_gateways[0], None)
+                return CouchbaseCloud([self.sync_gateways[0]], None)
 
             raise CblTestError("No Couchbase Servers available in config")
 
-        return CouchbaseCloud(self.sync_gateways[0], self.couchbase_servers[0])
+        return CouchbaseCloud([self.sync_gateways[0]], self.couchbase_servers[0])
--- a/client/src/cbltest/api/cloud.py
+++ b/client/src/cbltest/api/cloud.py
@@ -7,6 +7,7 @@ from opentelemetry.trace import get_tracer
 from cbltest.api.couchbaseserver import CouchbaseServer
 from cbltest.api.error import CblSyncGatewayBadResponseError, CblTestError
 from cbltest.api.syncgateway import PutDatabasePayload, SyncGateway
+from cbltest.api.syncgatewaycluster import SyncGatewayCluster
 from cbltest.assertions import _assert_not_null
 from cbltest.jsonhelper import _get_typed_required
 from cbltest.utils import _try_n_times
@@ -15,22 +16,41 @@ from cbltest.version import VERSION
 
 class CouchbaseCloud:
     """
-    A class that performs operations that require coordination between both Sync Gateway and Couchbase Server
+    A class that performs operations that require coordination between both Sync Gateway
+    and Couchbase Server.
     """
 
-    def __init__(self, sync_gateway: SyncGateway, server: CouchbaseServer | None):
-        self.__sync_gateway = sync_gateway
+    def __init__(
+        self,
+        sync_gateways: list[SyncGateway],
+        server: CouchbaseServer | None,
+    ):
+        if not sync_gateways:
+            raise CblTestError("At least one Sync Gateway must be provided")
+        self.__sync_gateways = sync_gateways
+        self.__sync_gateway_cluster = SyncGatewayCluster(sync_gateways)
+
         if server:
             self.__couchbase_server: CouchbaseServer = server
-        elif not sync_gateway.using_rosmar:
+        elif len(self.__sync_gateways) > 1:
+            raise CblTestError(
+                "Couchbase Server must be provided when configuring multiple Sync Gateway nodes"
+            )
+        elif not self.__sync_gateways[0].using_rosmar:
             raise CblTestError(
                 "Couchbase Server must be provided if Sync Gateway is not using Rosmar"
             )
         self.__tracer = get_tracer(__name__, VERSION)
 
     @property
-    def sync_gateway(self) -> SyncGateway:
-        return self.__sync_gateway
+    def sync_gateways(self) -> list[SyncGateway]:
+        """All Sync Gateway nodes managed by this Couchbase Cloud instance."""
+        return self.__sync_gateways
+
+    @property
+    def sync_gateway_cluster(self) -> SyncGatewayCluster:
+        """A `SyncGatewayCluster` view over all Sync Gateway nodes managed by this instance."""
+        return self.__sync_gateway_cluster
 
     @property
     def couchbase_server(self) -> CouchbaseServer:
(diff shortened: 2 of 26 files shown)


vipbhardwaj (approved):
LGTM

client/src/cbltest/api/cloud.py
```diff
@@ -15,22 +16,41 @@
...
 
-    def __init__(self, sync_gateway: SyncGateway, server: CouchbaseServer | None):
-        self.__sync_gateway = sync_gateway
+    def __init__(
+        self,
+        sync_gateways: list[SyncGateway],
+        server: CouchbaseServer | None,
+    ):
+        if not sync_gateways:
+            raise CblTestError("At least one Sync Gateway must be provided")
```
borrrden, line 29: In python does this mean "zero length"?  It's not listed as nullable so a null check here doesn't have a large return.

client/src/cbltest/api/cloud.py
```diff
@@ -41,7 +61,7 @@ def couchbase_server(self) -> CouchbaseServer:
         return self.__couchbase_server
 
     def _create_collections(self, db_payload: PutDatabasePayload) -> None:
-        if self.sync_gateway.using_rosmar:
+        if self.__sync_gateways[0].using_rosmar:
```
borrrden, line 64: In a multi sgw world, are we simply going to always use the first one, and is there any reason to believe that the second might be different?  They shouldn't be if part of the same cloud, I suppose.  They should all be part of the same "house" so to speak.

client/src/cbltest/api/cloud.py
```diff
@@ -117,21 +137,22 @@ async def configure_dataset(
...
                 if e.code != 412:
                     raise
 
                 current_span.add_event("Handle HTTP 412")
-                await self.__sync_gateway.delete_database(dataset_name)
+                await sg.delete_database(dataset_name)
                 await self.drop_bucket(db_payload.bucket)
-                await self.__sync_gateway.wait_for_no_databases(db_payload.bucket)
-                if not self.sync_gateway.using_rosmar:
+                await self.sync_gateway_cluster.wait_for_no_databases(db_payload.bucket)
```
borrrden, line 154: How does this wait_for_no_databases call work now that you changed this function name below?

borrrden: Nevermind it is in two places

client/src/cbltest/__init__.py
```diff
@@ -215,8 +215,8 @@ def simple_cloud(self) -> CouchbaseCloud:
 
         if len(self.couchbase_servers) == 0:
             if self.sync_gateways[0].using_rosmar:
-                return CouchbaseCloud(self.sync_gateways[0], None)
+                return CouchbaseCloud([self.sync_gateways[0]], None)
 
             raise CblTestError("No Couchbase Servers available in config")
 
-        return CouchbaseCloud(self.sync_gateways[0], self.couchbase_servers[0])
+        return CouchbaseCloud([self.sync_gateways[0]], self.couchbase_servers[0])
```
borrrden, line 222: Is the list actually longer than this?  If not you could just pas self.sync_gateways right?

client/src/cbltest/api/cloud.py
torcolvin: This means empty list, it's idiomatic to treat both identically in the time before types existed. This is defensive programming.

client/src/cbltest/__init__.py
torcolvin: I waffled on this - I think I would be inclined toward saying that the answer should be yes.

But I'd argue that this is a confusing limitation of the way topologies are defined.

Let's say you had a topology defined for this test https://github.com/couchbaselabs/couchbase-lite-tests/blob/b4c64c0feeca97db170432abbe8978a01d6346c1/tests/dev_e2e/test_replication_xdcr.py#L31-L55 but you are using it to run a test that called this `simple_cloud` function:

cbs1 -> couchbase://cbs1host
cbs2 -> couchbase://cbs2host
sg1 -> couchbase://cbs1host
sg2 -> couchbase://cbs2host

A test that uses `simple_cloud` would pass with this topology config if `CouchbaseCloud([self.sync_gateways[0]], self.couchbase_servers[0])` but not if `CouchbaseCloud([self.sync_gateways, self.couchbase_servers[0])` is true.

This problem might span outside this PR but if you assume that Sync Gateway as defined in a topology config will have a `server` hardcoded into it, then the topology abstraction is confusing.

Which of these are requirements for the topology configuration?

1. Couchbase Server and Sync Gateway are up and running at the addresses provided
2. Sync Gateway does not change the couchbase:// URL
3. Test Server / Edge Server are running when you start pytest
4. Any of these nodes stays running throughout the lifespan of the tests

I'm not suggesting that we go ahead and change topology configuration right now but my original inclination is to having a different mapping of topology configuration where you define Couchbase Clusters which contain 1 couchbase server [or 0 is rosmar] and N sync gateways.

It doesn't make sense to have sync gateways independent of a Couchbase Server because they are stateless, so even if you did restart a Sync Gateway to point to a different Couchbase Server, it doesn't have a lot of meaning.

I don't think we need to fix this right now, but I'm curious to get more clarity about this. For non Sync Gateway, Test Server or Edge Server, you actually want the ability to change the replication endpoints:

e.g. 
1. EdgeServer<->sg1
2. swap to EdgeServer<->sg2

We do have a test with a load balancer that does swap the hosts for the load balancer https://github.com/couchbaselabs/couchbase-lite-tests/blob/b4c64c0feeca97db170432abbe8978a01d6346c1/tests/dev_e2e/test_replication_xdcr.py#L217

I'm interested to have this discussion in more depth if only to clarify this for my Sync Gateway dev teammates who are more used to this type of topology setup.

client/src/cbltest/api/cloud.py
torcolvin: This is a conceptual problem that is confusing. Generally SyncGatewayCluster has wrapper functions for things that need to be cluster wide that basically delegate to each node in SyncGateway and collapse with asyncio.gather. So the functions have duplicated names. One idea is to often use single underscore methods in SyncGateway when they won't be called by test code.

client/src/cbltest/api/cloud.py
torcolvin: One concept that I was interested to implement was the concept of `round_robin_node` or `random_node` which I think would help in testing. This is not applicable if there's a load balancer in front of the nodes - though in that case this is an interesting problem where sometimes making a call to modify one node would not be reflected on all nodes until the polling window collapses.

Any request call can happen on any node, like this one. I think this would be a good time for `random_node` here, though that makes the tests mildly non deterministic.

For a few operations - anything that modifies a db - you want to wait for that change to get propagated to all nodes. It is those functions alone that would need to implemented in `SyncGatewayCluster`. The idea behind this is to avoid any sleep statements which are subject to races.

PR #447, opened 2026-07-20
Add periodic Sync Gateway scheduler that triggers sgw-functional only for new builds

sanjivanipatrax:
This PR adds a periodic scheduler that auto-runs the SGW functional QE pipeline, resolving the latest Sync Gateway build and triggering `sgw-functional` only when that build hasn't run yet.

**Features added:**
- New `jenkins/pipelines/QE/sgw-main/Jenkinsfile` - a scheduler pinned to `mac-mini-new`, running daily at 4 AM IST.
- Resolves the latest SGW version + build (dedup key) and the latest `couchbase-lite-ios` client version + build via Proget, and triggers `sgw-functional` with both pinned as `<version>-<build>`.
- Per-build dedup on the SGW build via `~/.jenkins_build_info/<JOB_NAME>/sgw.json`; `FORCE_TRIGGER` overrides.
- `SGW_VERSION/CBL_VERSION` params for manual runs; build display name shows the resolved versions.

Verified against the job pointed at this branch.

### DIFF
--- a/jenkins/pipelines/QE/sgw-main/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw-main/Jenkinsfile
@@ -33,7 +33,7 @@ pipeline {
         booleanParam(name: 'FORCE_TRIGGER', defaultValue: false, description: 'Force triggering the SGW test job even if this SGW version+build already ran')
     }
     options {
-        timeout(time: 30, unit: 'MINUTES')
+        timeout(time: 1, unit: 'HOURS')
         disableConcurrentBuilds()
     }
     triggers {
--- a/jenkins/pipelines/QE/sgw-main/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw-main/Jenkinsfile
@@ -37,7 +37,8 @@ pipeline {
         disableConcurrentBuilds()
     }
     triggers {
-        cron('@daily')
+        cron('''TZ=Asia/Kolkata
+0 4 * * *''')
     }
     stages {
         stage('Resolve and Trigger') {
--- a/jenkins/pipelines/QE/sgw-main/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw-main/Jenkinsfile
@@ -34,6 +34,7 @@ pipeline {
     }
     options {
         timeout(time: 30, unit: 'MINUTES')
+        disableConcurrentBuilds()
     }
     triggers {
         cron('@daily')
--- a/jenkins/pipelines/QE/sgw-main/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw-main/Jenkinsfile
@@ -1,0 +1,95 @@
+def SGW_PRODUCT = "sync-gateway"          // dedup key: the thing under test
+def CBL_PRODUCT = "couchbase-lite-ios"    // iOS test-server client, passed through
+
+// A fully-qualified "<version>-<build>" param is used as-is; otherwise the latest
+// build is looked up via ProGet get_version.
+def resolveVersionBuild(proget, String product, String param, String label) {
+    def version = proget.resolveProgetVersion(product, param, label)
+    def buildNo
+    if (version.contains('-')) {
+        def parts = version.split('-', 2)
+        version = parts[0]
+        buildNo = parts[1]
+    } else {
+        def buildUrl = "http://proget.build.couchbase.com:8080/api/get_version?product=${product}&version=${version}"
+        buildNo = sh(
+            script: "curl -sf '${buildUrl}' | python3 -c 'import json,sys; print(json.load(sys.stdin).get(\"BuildNumber\",\"\"))'",
+            returnStdout: true
+        ).trim()
+    }
+    if (!buildNo) { error "Could not resolve build number for ${product} ${version}" }
+    return [version: version, build: buildNo]
+}
+
+pipeline {
+    // Pinned: dedup state lives on this node's filesystem (~/.jenkins_build_info).
+    agent { label 'mac-mini-new' }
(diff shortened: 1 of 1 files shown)


jenkins/pipelines/QE/sgw-main/Jenkinsfile
```diff
@@ -0,0 +1,97 @@
...
+0 4 * * *''')
+    }
+    stages {
+        stage('Resolve and Trigger') {
+            steps {
+                script {
+                    def proget = load 'jenkins/pipelines/shared/resolveProgetVersion.groovy'
+
+                    def sgw = resolveVersionBuild(proget, SGW_PRODUCT, params.SGW_VERSION, 'SGW_VERSION')
+                    def cbl = resolveVersionBuild(proget, CBL_PRODUCT, params.CBL_VERSION, 'CBL_VERSION')
```
borrrden, line 50: As discussed yesterday I'm not sure that getting the latest prerelease build for CBL AND SGW in the same job is something that is worth running regularly.

borrrden (changes requested):

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: As discussed yesterday, `sgw-functional` already fetches the latest prerelease builds for both CBL and SGW. This scheduler follows the same design and passes those builds to `sgw-functional`, so I think we’d need to revisit that design as well if we don’t want to regularly run against the latest prerelease builds for both.

vipbhardwaj: I think this can be handled by the jenkins job passing a build intended for the test. This can just be a fail-safe when nothing concrete has been passed.

jenkins/pipelines/QE/sgw-main/Jenkinsfile
```diff
@@ -0,0 +1,97 @@
+def SGW_PRODUCT = "sync-gateway"          // dedup key: the thing under test
+def CBL_PRODUCT = "couchbase-lite-ios"    // iOS test-server client, passed through
+
+// A fully-qualified "<version>-<build>" param is used as-is; otherwise the latest
+// build is looked up via ProGet get_version.
+def resolveVersionBuild(proget, String product, String param, String label) {
```
vipbhardwaj, line 6: This is redundant code, it can be imported.

vipbhardwaj (changes requested):

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: @vipbhardwaj Could you clarify what you mean by a “build intended for the test”? I’m not sure I follow the proposed flow.

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: I looked for an existing implementation before writing this and couldn’t find one. Could you point me to the one you’re referring to?

vipbhardwaj: [https://github.com/couchbaselabs/couchbase-lite-tests/blob/main/jenkins/pipelines/QE/sgw/Jenkinsfile](https://github.com/couchbaselabs/couchbase-lite-tests/blob/main/jenkins/pipelines/QE/sgw/Jenkinsfile)

jenkins/pipelines/QE/sgw-main/Jenkinsfile
vipbhardwaj: Don't use `resolveVersionBuild` for both, instead let one of them to be specified by the user specifically, that way we would prevent the current behaviour of : 
> latest prerelease build for CBL AND SGW in the same job

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: `sgw/Jenkinsfile` (lines 16–18) just calls `resolveProgetVersion` and stores the combined string - it never splits version from build or does the BuildNumber lookup. `sgw-main` needs them separate for the dedup key (`buildInfo[sgw.version] == sgw.build`), which is exactly what `resolveVersionBuild` adds.

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: This is a periodic scheduler that’s intended to run automatically. I’m not sure how requiring user input would work in that model.

jenkins/pipelines/QE/sgw-main/Jenkinsfile
sanjivanipatrax: Going forward, it would help if we could first confirm that the existing implementation provides the same behaviour before marking code as redundant.

jenkins/pipelines/QE/sgw-main/Jenkinsfile
borrrden: I think probably the best thing to do would be to find a way to resolve the difference between the two functions.  Splitting the build and version is easy enough after it gets returned, and perhaps the function could have a prerelease argument you could pass so that it checks for either the latest release or latest prerelease.  Out of scope of this PR though.

borrrden (approved):

borrrden:
I want to address both @vipbhardwaj and my comments in a separate PR but for now this will be functionally ok.

PR #450, opened 2026-07-22
Modify device used for c-ios QE tests

barkha06:
Modified device from iPhone 16e ( seeing a failed to allocate RSD disk error which seems to be resolved only by unpairing and pairing the device back) to iPad Pro.


borrrden:
You can push changes like this without a review

borrrden (approved):

PR #454, opened 2026-07-24
CBL-8648: gate edge_server JWT tests with min_edge_servers marker & Separate dev_e2e into cbl/ and edge_server/ suites

nitishgupta25:
Add min_edge_servers / min_sync_gateways / min_couchbase_servers markers to TestJWTSimple and TestJWTReplication so they skip (instead of crashing with IndexError) on dev_e2e runs that have no edge servers. Also fix the stale sgw-auto-cleanup comment.

Move the CBL test files into tests/dev_e2e/cbl/ so the CBL and Edge Server suites are separate siblings (edge_server ships on its own version cadence). Point the per-platform dev_e2e CI scripts at cbl/, and update docs (AGENTS.md, README) to the new layout.


nitishgupta25:
Heads up, my PR #454 (edge_server JWT test markers) is failing python-format/python-lint, but it's the repo-wide ruff 0.16 issue

borrrden (changes requested):
You missed the powershell scripts that are meant to run on Windows.  Looks like you only edited the shell script ones

borrrden:
This is going to end up ideologically conflicting with #457 and I think before either can be approved we need to get a design document going about how we organize our tests.

vipbhardwaj:
> This is going to end up ideologically conflicting with #457 and I think before either can be approved we need to get a design document going about how we organize our tests.

Having a separate sub-folder for CBL right now, its fine to merge this right now. My PR can have a directory-based design discussion later on. Right now my PR has further discussions so we do not need to block this 👍🏻

nitishgupta25:
Per the agreement to use marker-based filtering in a single folder, closing this- the cbl/ folder split is no longer the direction. min_* markers are already handled in #458

PR #457, opened 2026-07-29
Combined run for all SGW tests - backend

vipbhardwaj:
[CBG-5648](https://jira.issues.couchbase.com/browse/CBG-5648)

### NIGHTLY PR
Implementation : 
- The conftest.py uses a script to mark nightly tests, every test inside /tests/dev_e2e is marked, even Edge Server
- c_macos is used instead of previous swift-ios
- Since the test runs at parent dir level, ie, `/tests/**`, hence the artifacts are stored at parent dir level too

To Be Addressed later : 
- The edge-server tests being marked nightly as well, and hence ignored via the pytest command later
- An XDCR test is currently ignored too since the topology uses a single 2 node CBS cluster, I was looking at a workaround for this, want to file another ticket + PR for that. OUT OF SCOPE for this PR.

Any other issue related to the implementation, eg @torcolvin a workaround to leave edge-server dir out naturally (considering its okay to leave it out of nighly??) can be used.
[Link for the test run for verification](http://jenkins.mobiledev.couchbase.com/view/QE/job/sgw-functional/180/). Ignore the failures, they are because of an old PR **merged to main**, I believe #442, which I will fix soon.

[CBG-5648]: https://couchbasecloud.atlassian.net/browse/CBG-5648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/jenkins/pipelines/QE/sgw/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw/Jenkinsfile
@@ -24,7 +24,7 @@ pipeline {
             steps {
                 build job: 'prebuild-test-server',
                 parameters: [
-                    string(name: 'TS_PLATFORM', value: 'swift_ios'),
+                    string(name: 'TS_PLATFORM', value: 'c_macos'),
                     string(name: 'CBL_VERSION', value: env.CBL_VERSION),
                 ],
                 wait: true,
@@ -54,8 +54,8 @@ pipeline {
                     // surfaced in Jenkins even if teardown fails.
                     script {
                         try {
-                            junit testResults: 'tests/QE/junit_result.xml', allowEmptyResults: true
-                            archiveArtifacts artifacts: 'tests/QE/junit_result.xml', fingerprint: true, allowEmptyArchive: true
+                            junit testResults: 'tests/junit_result.xml', allowEmptyResults: true
+                            archiveArtifacts artifacts: 'tests/junit_result.xml', fingerprint: true, allowEmptyArchive: true
                         } catch (err) {
                             echo "WARNING: Failed to publish/archive JUnit XML: ${err}"
                         }
@@ -66,8 +66,9 @@ pipeline {
                         sh "jenkins/pipelines/QE/sgw/teardown.sh"
                     }
                     // Archived after teardown because teardown's move_artifacts
-                    // is what populates tests/QE/sgw/.
-                    archiveArtifacts artifacts: 'tests/QE/sgw/**/*', fingerprint: true, allowEmptyArchive: true
+                    // is what populates tests/sgw/ (the whole tests/ tree runs
+                    // in one session, so artifacts live at the tests/ root).
+                    archiveArtifacts artifacts: 'tests/sgw/**/*', fingerprint: true, allowEmptyArchive: true
                     echo "=== SGW Test Teardown Complete"
                 }
             }
--- a/jenkins/pipelines/QE/sgw/Jenkinsfile
+++ b/jenkins/pipelines/QE/sgw/Jenkinsfile
@@ -54,8 +54,8 @@ pipeline {
                     // surfaced in Jenkins even if teardown fails.
                     script {
                         try {
-                            junit testResults: 'tests/junit_result.xml', allowEmptyResults: true
-                            archiveArtifacts artifacts: 'tests/junit_result.xml', fingerprint: true, allowEmptyArchive: true
+                            junit testResults: '*/junit_result.xml', allowEmptyResults: true
+                            archiveArtifacts artifacts: '*/junit_result.xml', fingerprint: true, allowEmptyArchive: true
                         } catch (err) {
                             echo "WARNING: Failed to publish/archive JUnit XML: ${err}"
                         }
--- a/jenkins/pipelines/QE/sgw/test.sh
+++ b/jenkins/pipelines/QE/sgw/test.sh
@@ -27,11 +27,11 @@ for arg in "$@"; do
 done
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-source $SCRIPT_DIR/../../shared/config.sh
+source "$SCRIPT_DIR"/../../shared/config.sh
 
 echo "Setup backend..."
-pushd $AWS_ENVIRONMENT_DIR > /dev/null
-uv run $SCRIPT_DIR/setup_test.py $CBL_VERSION $SGW_VERSION
+pushd "$AWS_ENVIRONMENT_DIR" > /dev/null
+uv run "$SCRIPT_DIR"/setup_test.py "$CBL_VERSION" "$SGW_VERSION"
 popd > /dev/null
 
 # Exit early if setup-only mode
@@ -57,7 +57,7 @@ fi
 # and tests/QE/test_multipeer.py share a basename and would otherwise collide
 # at import time (pytest imports every collected module before -m deselection).
 echo "Run tests..."
-pushd $TESTS_DIR > /dev/null
+pushd "$TESTS_DIR" > /dev/null
 uv run pytest -v --no-header -W ignore::DeprecationWarning \
     --config QE/config.json \
     --import-mode=importlib \
@@ -65,4 +65,5 @@ uv run pytest -v --no-header -W ignore::DeprecationWarning \
     --ignore=QE/edge_server \
     --ignore=dev_e2e/edge_server \
     --ignore=dev_e2e/test_replication_xdcr.py \
-    --sgcollect-on-test-failure
+    --sgcollect-on-test-failure \
+    dev_e2e QE
--- a/jenkins/pipelines/QE/sgw/test.sh
+++ b/jenkins/pipelines/QE/sgw/test.sh
@@ -40,30 +40,17 @@ if [ "$SETUP_ONLY" = true ]; then
     exit 0
 fi
 
-# Run Tests :
-# Run the whole tests/ tree (both QE and dev_e2e) in a single session and let
-# markers pick the SGW-relevant tests:
-#   * "sgw"  -> QE tests already tagged for Sync Gateway.
-#   * "nightly" -> every dev_e2e test (auto-tagged in tests/dev_e2e/conftest.py).
-#   * "and min_sync_gateways" -> drop tests that never touch SGW (e.g. the
-#     P2P-only tests/dev_e2e/test_multipeer.py, which has no min_sync_gateways).
-# The edge_server sub-suites are out of scope for the nightly SGW run, so they
-# are excluded explicitly rather than by marker.
 # test_replication_xdcr.py needs two separate Couchbase Server clusters, which
 # this single-cluster topology does not provide; it is deferred until the
-# multi-cluster topology is sorted out (tracked separately) rather than left to
-# fail every night.
+# multi-cluster topology is sorted out rather than left to fail every night.
+
 # --import-mode=importlib is required because tests/dev_e2e/test_multipeer.py
 # and tests/QE/test_multipeer.py share a basename and would otherwise collide
-# at import time (pytest imports every collected module before -m deselection).
+# at import time can be removed after: https://jira.issues.couchbase.com/browse/CBL-8685.
 echo "Run tests..."
 pushd "$TESTS_DIR" > /dev/null
-uv run pytest -v --no-header \
+uv run pytest -v --no-header -m sgw \
     --config QE/config.json \
     --import-mode=importlib \
-    -m "(sgw or nightly) and min_sync_gateways" \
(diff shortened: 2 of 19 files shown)


tests/dev_e2e/conftest.py
```diff
@@ -4,6 +4,25 @@
...
+#
+# This hook is registered by every conftest in the collection tree, so it must
+# only mark the items that live under *this* directory — otherwise a combined
+# `tests/`-level run would also tag the QE suite.
+def pytest_collection_modifyitems(
+    config: pytest.Config, items: list[pytest.Item]
+) -> None:
+    for item in items:
+        if _THIS_DIR in Path(str(item.fspath)).resolve().parents:
+            item.add_marker(pytest.mark.nightly)
```
borrrden: I think @torcolvin and I might need to argue about this for a while because this feels counterproductive.  Why do we just blanket mark every single test in this folder as nightly when some of them clearly are not meant to be run nightly.   This combined with the fact that "nightly" should not mean "SGW nightly" in the first place.  CBL tests are in here too and adding in a hodge podge of ignore statements to whittle down to the ones you ACTUALLY want seems more like a symptom of improper organization.  

We really need to come up with a plan of organization before trying to do something like this considering you have tests in here that don't make sense for sync gateway, but happen to use sync gateway (prime example is the test_query_consistency suite) to populate data for a CBL test.  

Plus, with this method every new test that gets added here is magically and silently a nightly test.  I think this may all be premature and what we actually need to do is sit down and logically organize the tests based on product and category of test.  I think this would mostly land at the class level anyway so we are only talking about a handful of added tags.

pyproject.toml
```diff
@@ -57,12 +57,17 @@ root = [".", "client/src", "tests"]
...
+# uses so the two test_multipeer.py modules don't collide at collection.
+pythonpath = ["tests", "tests/dev_e2e"]
 filterwarnings = [
   "ignore:Class property max_ttl is deprecated.*:couchbase.logic.supportability.CouchbaseDeprecationWarning",
 ]
 markers = [
   "sgw: tests that focus on Sync Gateway functionality",
   "upg_sgw: tests that focus on Sync Gateway upgrade scenario",
   "cbl: tests that focus on Couchbase Lite functionality",
+  "nightly: tests run by the nightly SGW job (auto-applied to all tests/dev_e2e tests via tests/dev_e2e/conftest.py)",
```
borrrden: This irks me a bit because I imagine we are going to use this marker for more than just sgw tests.  I'm still of the opinion that we should be tagging this appropriately so we can simply say (sgw and nightly) for the filter.  If we get that correct we can actually collapse all the tests into a single directory.

jenkins/pipelines/QE/sgw/Jenkinsfile
```diff
@@ -54,8 +54,8 @@ pipeline {
                     // surfaced in Jenkins even if teardown fails.
                     script {
                         try {
-                            junit testResults: 'tests/QE/junit_result.xml', allowEmptyResults: true
-                            archiveArtifacts artifacts: 'tests/QE/junit_result.xml', fingerprint: true, allowEmptyArchive: true
+                            junit testResults: 'tests/junit_result.xml', allowEmptyResults: true
+                            archiveArtifacts artifacts: 'tests/junit_result.xml', fingerprint: true, allowEmptyArchive: true
```
torcolvin:
```suggestion
                            junit testResults: '*/junit_result.xml', allowEmptyResults: true
                            archiveArtifacts artifacts: '*/junit_result.xml', fingerprint: true, allowEmptyArchive: true
```

might be more foward thinking here if we decide on a restructure or substructure.

tests/dev_e2e/conftest.py
torcolvin: I think we can avoid coming up with a bigger plan and drop this complexity.

From a discussion in the QE weekly meeting:

1. We can run tests that don't really use sync gateway because they don't take much time.
2. We just need to skip some tests right now that are missing min_test_servers https://github.com/couchbaselabs/couchbase-lite-tests/pull/458/

Tests will be opt-out automatically pending fixing test markers:

1. any test that requires edge-server
2. we could drop multipeer tests by only deploying one test server and relying on markers

Then we can put off adding a marker `nightly` which I don't think either Jim or I like but didn't communicate well about. I think _when_ we need a marker for something, this should be an opt-out marker rather than an opt-in marker, something like "test_slower_than_5m" or "requires_extra_resources" and doesn't describe when we run it but how we run it. I think we can enable all of the Sync Gateway tests for now, and then you can report a breakdown of how long they take to run. We can merge this PR before deciding on how long.

tl;dr let's just drop out the nightly markers from this PR.

jenkins/pipelines/QE/sgw/test.sh
```diff
@@ -41,7 +41,29 @@ if [ "$SETUP_ONLY" = true ]; then
...
+# fail every night.
+# --import-mode=importlib is required because tests/dev_e2e/test_multipeer.py
+# and tests/QE/test_multipeer.py share a basename and would otherwise collide
+# at import time (pytest imports every collected module before -m deselection).
 echo "Run tests..."
-pushd $QE_TESTS_DIR > /dev/null
-uv run pytest -v --no-header -W ignore::DeprecationWarning --config config.json -m sgw \
-    --sgcollect-on-test-failure
+pushd "$TESTS_DIR" > /dev/null
+uv run pytest -v --no-header -W ignore::DeprecationWarning \
```
torcolvin:
```suggestion
uv run pytest -v --no-header \
```

This is just boilerplate from before, but I've made sure there are no DeprecationWarnings. I think we want to know if they happen, and if there are unsuppressable warnings, we can use `warnings.catch_warnings` in the code or pyproject.toml to suppress them. I'm nearly certain there are no warnings though.

jenkins/pipelines/QE/sgw/test.sh
```diff
@@ -41,7 +41,29 @@ if [ "$SETUP_ONLY" = true ]; then
...
+    --config QE/config.json \
+    --import-mode=importlib \
```
torcolvin: This is trying to work around the fact that https://jira.issues.couchbase.com/browse/CBL-8685 isn't working yet. I'm fine if https://jira.issues.couchbase.com/browse/CBL-8685 doesn't get merged quickly to leave this but drop a comment in this line and on that CBL ticket to say remove this import-mode=importlib.

jenkins/pipelines/QE/sgw/test.sh
```diff
@@ -41,7 +41,29 @@ if [ "$SETUP_ONLY" = true ]; then
...
+    -m "(sgw or nightly) and min_sync_gateways" \
+    --ignore=QE/edge_server \
+    --ignore=dev_e2e/edge_server \
```
torcolvin: I have a strong preference to not include these lines (though they are harmless) and merge https://github.com/couchbaselabs/couchbase-lite-tests/pull/458/ before this.

jenkins/pipelines/QE/sgw/test.sh
```diff
@@ -41,7 +41,29 @@ if [ "$SETUP_ONLY" = true ]; then
...
+    --ignore=dev_e2e/test_replication_xdcr.py \
```
torcolvin, line 55: Why does this need to be skipped?

jenkins/pipelines/QE/sgw/test.sh
```diff
@@ -41,7 +41,29 @@ if [ "$SETUP_ONLY" = true ]; then
...
+    --sgcollect-on-test-failure \
+    dev_e2e QE
```
torcolvin: you can run without a test directory argument and it will run in cwd

torcolvin (commented):
I think we can drop the conftest.py and just run all the tests.

If there are things that are blocking this PR from merging after like:

https://github.com/couchbaselabs/couchbase-lite-tests/pull/458/
or
https://jira.issues.couchbase.com/browse/CBL-8685

just add a comment to say to remove these lines when the jira ticket or PR is merged, and note on the other PR that they should remove those lines.

jenkins/pipelines/QE/sgw/Jenkinsfile
vipbhardwaj: Makes sense.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: Okay.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: It uses 2-CBS clusters which would need a SGW topology restructure as well since currently it has a single 2-node cluster. Hence I decided to address this in an upcoming PR I've been working on.
I have been thinking how 2 single-node clusters would affect the approach, vs, a 2-node cluster (like now) with another SINGLE-NODE cluster (just for this test, nothing else). Provisioning a third node would add setup/cleanup overhead.
And I'm thinking that would the approach of 2 single node clusters affect the current tests setup on 1 cluster with 2 nodes.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: Yes I just wanted to order it. Right now because of auto-ordering-due-to-capital-QE, it was running QE first, which I don't mind. The initial intention was to find the root cause of #460 .

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: Okay.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: #458 is merged so I'l remove these lines. I think the markers added via that will ignore */edge_server.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: For now I am not covering this one test from `dev_e2e` so that it doesn't fail due unnecessarily.

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: I'll remove it now since the RootCause for that failure has been verified, #460 will have a fix after random test order run as well.

vipbhardwaj:
@torcolvin I have addressed your comments and fixed everything except adding a `dev_e2e/edge_server` because that is needed till `--ignore=dev_e2e/edge_server` is also kept, both can together be removed once the [ticket](https://jira.issues.couchbase.com/browse/CBL-8685) resolves as the comment in the code says.
Do let me know of any other short-falls

jenkins/pipelines/QE/sgw/test.sh
torcolvin: Gotcha! this makes sense. I think this merits a separate jira ticket, let me file one!

pyproject.toml
torcolvin: Can you drop this line of code since we aren't using it?

pyproject.toml
```diff
@@ -57,12 +57,17 @@ root = [".", "client/src", "tests"]
 
 [tool.pytest.ini_options]
 asyncio_default_fixture_loop_scope = "session"
-pythonpath = ["tests"]
+# "tests" puts tests/shared on the path; "tests/dev_e2e" keeps the bare
+# sibling import in test_replication_filter.py (from test_replication_filter_data
+# import ...) working under --import-mode=importlib, which the nightly SGW job
+# uses so the two test_multipeer.py modules don't collide at collection.
+pythonpath = ["tests", "tests/dev_e2e"]
```
torcolvin, line 64: I think you can leave this as is and just to --import-mode=importlib on this test run with a comment to remove this in https://jira.issues.couchbase.com/browse/CBL-8685. I think we can get a fix pretty quickly for that ticket but I don't want to hold this up.

tests/dev_e2e/conftest.py
torcolvin: Let's drop the marker from pyproject.toml as well and then this is done!

jenkins/pipelines/QE/sgw/test.sh
torcolvin: I think just mark https://jira.issues.couchbase.com/browse/CBL-8686 why this exists, and then this is fine!

pyproject.toml
vipbhardwaj: Done

jenkins/pipelines/QE/sgw/test.sh
vipbhardwaj: Done

PR #459, opened 2026-07-30
CBG-5636 allow running Couchbase Server from cbdinocluster

torcolvin:
CBG-5636 allow running Couchbase Server from cbdinocluster

- drop the requirement to have couchbase on localhost
- move all code into ./start_local.py

relates to https://github.com/couchbase/sync_gateway/pull/8492 - it is possible I should include how to start cbs here as well.


torcolvin (commented):
I'm going to merge this without as much review while Jim is out because it only affects Sync Gateway test code! Feel free to retrospectively comment.

PR #460, opened 2026-07-30
TDK - CBS port mapping fix

vipbhardwaj:
[CBG-5656](https://jira.issues.couchbase.com/browse/CBG-5656)

Addressing fix for failing SGW tests.
After a lot of "doubting" different reasons, I ran with one change. Seems to have fixed everything. Not sure how this for affecting every next test after this as well. I expected the array un-packing would only make this test fail, and there might be some other issue affecting every next test as well. But seems to be the case that this was affecting everything else coming after this. Ran two test runs, TEST LINKS IN THE TICKET ABOVE.

Update on the description : 
The port mapping was not complete in `couchbasesever.py` and `_set_alternate_address` in the test was a defensive repair of cluster addressing that a prior test (TestMultipleServers via add_node) had broken on purpose. With external KV addressing now guaranteed by provisioning (initial) + fixed add_node (after any re-add), the alternate addresses are complete for the entire run. So by the time ISGR reaches its setup, KV is already reachable — and ISGR does need KV (its cbs.create_collections() does a KV readiness get()), which now just works.

[CBG-5656]: https://couchbasecloud.atlassian.net/browse/CBG-5656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/src/cbltest/api/couchbaseserver.py
+++ b/client/src/cbltest/api/couchbaseserver.py
@@ -874,7 +874,17 @@ class CouchbaseServer:
                 try:
                     self.__http_session.put(
                         f"http://{node_to_add.__hostname}:8091/node/controller/setupAlternateAddresses/external",
-                        data={"hostname": node_to_add.__hostname, "mgmt": "8091"},
+                        data={
+                            "hostname": node_to_add.__hostname,
+                            "kv": "11210",
+                            "kvSSL": "11207",
+                            "mgmt": "8091",
+                            "mgmtSSL": "18091",
+                            "capi": "8092",
+                            "capiSSL": "18092",
+                            "n1ql": "8093",
+                            "n1qlSSL": "18093",
+                        },
                     ).raise_for_status()
                     span.add_event(
                         "alternate_address_set",
--- a/client/src/cbltest/api/couchbaseserver.py
+++ b/client/src/cbltest/api/couchbaseserver.py
@@ -874,17 +874,7 @@ class CouchbaseServer:
                 try:
                     self.__http_session.put(
                         f"http://{node_to_add.__hostname}:8091/node/controller/setupAlternateAddresses/external",
-                        data={
-                            "hostname": node_to_add.__hostname,
-                            "kv": "11210",
-                            "kvSSL": "11207",
-                            "mgmt": "8091",
-                            "mgmtSSL": "18091",
-                            "capi": "8092",
-                            "capiSSL": "18092",
-                            "n1ql": "8093",
-                            "n1qlSSL": "18093",
-                        },
+                        data={"hostname": node_to_add.__hostname, "mgmt": "8091"},
                     ).raise_for_status()
                     span.add_event(
                         "alternate_address_set",
--- a/tests/QE/test_multiple_servers.py
+++ b/tests/QE/test_multiple_servers.py
@@ -326,7 +326,7 @@ class TestISGRCollectionMapping(CBLTestClass):
     ) -> None:
         sg_cluster = SyncGatewayCluster(cblpytest.sync_gateways)
         cbs = cblpytest.couchbase_servers[0]
-        sg1, sg2, sg3 = sg_cluster.sync_gateways[:2]
+        sg1, sg2, sg3 = sg_cluster.sync_gateways[:3]
         bucket1, bucket2, bucket3 = "isgr-bucket1", "isgr-bucket2", "isgr-bucket3"
         sg_db1, sg_db2, sg_db3 = "db1", "db2", "db3"
         b1_collections = ["collection1", "collection2", "collection3"]
--- a/tests/QE/test_multiple_servers.py
+++ b/tests/QE/test_multiple_servers.py
@@ -66,39 +66,82 @@ async def _setup_database_and_user(
     return await sg.create_user_client(sg_db, user_name, user_password, channels)
 
 
-def _readvertise_external_addresses(cbs_servers: list) -> None:
+def _snapshot_cluster(cbs_servers: list) -> tuple[dict[str, str], set[str]]:
     """
-    Re-advertise the external alternate address on every CBS node so the external
-    test agent's SDK can reach all services (incl. KV). A node that was
-    ejected/re-added or failed-over/recovered can lose its alternate address,
-    leaving it unreachable from outside the VPC; this restores it. Setting only
-    the hostname makes Couchbase Server auto-map all standard service ports --
-    the same thing provisioning's `couchbase-cli setting-alternate-address` does.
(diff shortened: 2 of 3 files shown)


client/src/cbltest/api/couchbaseserver.py
```diff
@@ -874,7 +874,17 @@ def do_add_node() -> None:
...
+                            "hostname": node_to_add.__hostname,
+                            "kv": "11210",
+                            "kvSSL": "11207",
+                            "mgmt": "8091",
+                            "mgmtSSL": "18091",
+                            "capi": "8092",
+                            "capiSSL": "18092",
+                            "n1ql": "8093",
+                            "n1qlSSL": "18093",
+                        },
```
torcolvin: I don't see why this is necessary to add the port numbers?

I was looking at the source of couchbase-cli setting-alternate-address which calls this code https://github.com/couchbase/couchbase-cli/blob/master/cluster_manager.py#L2982

This is why is used in the setup script https://github.com/couchbaselabs/couchbase-lite-tests/blob/6f66821c16c3ab1f8553617e2d13ce4497b2510c/environment/aws/server_setup/configure-node.sh#L89

I think you can drop this line.

tests/QE/test_multiple_servers.py
```diff
@@ -334,9 +313,6 @@ async def test_isgr_explicit_collection_mapping(
         b3_collections = ["collection6", "collection7", "collection8", "collection9"]
         num_docs = 3
 
-        self.mark_test_step("Clean up and setup test environment")
-        _set_alternate_addresses(cblpytest.couchbase_servers)
```
torcolvin, line 338: I think there's something here that might be relevant - I suspect this was trying to clean up from this test https://github.com/couchbaselabs/couchbase-lite-tests/blob/6f66821c16c3ab1f8553617e2d13ce4497b2510c/tests/QE/test_multiple_servers.py#L241 which caused a failover. I think that removing this line alone might be "sufficient" to fix the test.

I think what you want to do to fix this is to create a fixture that analyzes the state of the couchbase server cluster at the start of the test and restore to the state of the tests.

Can you reproduce by just running test_server_goes_down_sanity and test_isgr_explicit_collection_mapping together?

Look at all the tests in test_multiple_servers.py and also run the tests in this file with pytest-random-order plugin to make sure that you do the cleanup correctly.

vipbhardwaj: Yes I thought of adding a fixture too but the `add_node` was already there with the `data` param being set, which runs everytime when a AWS-CBS is setup. Hence this added minimal change

client/src/cbltest/api/couchbaseserver.py
vipbhardwaj: Okay so basically I can use `couchbase_cli` based `setting-alternate-address` to set addresses instead of the current method?

Or can you expand on how couchbase-cli can replace the current func `add_node` logic since I didn't understand properly.

tests/QE/test_multiple_servers.py
torcolvin: I don't think that data parameter has any affect because those are the default values - what I am suggesting is that this might be a coincidence for why this test starting passing, or it is just related to not calling _set_alternate_addresses on a cluster without re-joining all of the nodes

vipbhardwaj: You're right. Verified by using `--random-order` flag twice , both pass : 
```
test_multiple_servers.py::TestISGRCollectionMapping::test_isgr_explicit_collection_mapping PASSED       [ 33%]
test_multiple_servers.py::TestMultipleServers::test_server_goes_down_sanity PASSED                      [ 66%]
test_multiple_servers.py::TestMultipleServers::test_rebalance_sanity PASSED                             [100%]
```
and
```
test_multiple_servers.py::TestMultipleServers::test_rebalance_sanity PASSED                             [ 33%]
test_multiple_servers.py::TestMultipleServers::test_server_goes_down_sanity PASSED                      [ 66%]
test_multiple_servers.py::TestISGRCollectionMapping::test_isgr_explicit_collection_mapping PASSED       [100%]
```

client/src/cbltest/api/couchbaseserver.py
vipbhardwaj: I've removed this too. Resolving the conversation

tests/QE/test_multiple_servers.py
```diff
@@ -87,6 +66,42 @@ async def _setup_database_and_user(
...
+    session.auth = ("Administrator", "password")
+    for cbs_node in cbs_servers:
+        session.put(
+            f"http://{cbs_node.hostname}:8091/node/controller/setupAlternateAddresses/external",
+            data={"hostname": cbs_node.hostname},
+        )
+
+
+@pytest.fixture
+def restore_cbs_cluster(cblpytest: CBLPyTest):
```
torcolvin, line 129: I wonder if you can do something in this test fixture that does:

1. grab the alternate address + node status of all the couchbase server nodes before the test runs
2. yields
3. force restoring it in teardown

You don't just have to setup alternate addresses, you have to rejoin the nodes as well if they were ejected.

vipbhardwaj: Done

vipbhardwaj:
Ran a test against latest commit, [passed](http://jenkins.mobiledev.couchbase.com/job/sgw-functional/196/)  ✅

torcolvin:
I don't feel 100% comfortable with this PR because the snapshot of the cluster expects `cbs_servers[0]` to always exist, but you can eject the first node.

I've been tooling around with some changes that will always snapshot the whole state of `cbs_servers` for the following cases: alternateAddresses, port mapping, node ejection and making sure to restore to the previous state. In that case we can use a generic fixture to prevent this from the future.

There's some other stuff about this PR - it hardcodes the CBS credentials.

I think I'd feel most comfortable if we put a temporary skip `pytest.mark.skip("test does not clean up after a rebalance CBS-5656")` on tests which rebalance CBS - `TestMultipleServers`. Let's create a separate PR for this so greenboard will run green but we can keep this PR open to come to a long term fix.

Can you set up a meeting with me and @RIT3shSapata to go over the fix I have? I didn't quite finish this yesterday but I should have more time/tokens to finish this prototype. Then I can discuss all the things we might have to preserve / restore and some clever-ish ways to make the code idempotent. This is easier for me and I think will teach y'all more about this to do pair programing for this.

torcolvin:
Closing this ticket, we are going to go for a different approach of a central fixture to restore a couchbase server and sync gateway setup more generally.

